PR #3, opened 2015-01-22
License cleanup

beorn7:
@brian-brazil I just did the same as for the other repos. Minus "developed at SoundCloud". Feel free to change at will.


brian-brazil:
:+1:

PR #6, opened 2015-03-13
Allow filtering of metrics by dimension values

fstehle:
In order to avoid fetching all metrics for one Namespace (eg. `AWS/ELB`), I would like to introduce the possibility to filter metrics, by a metric dimension value.
That allows to have the following config to only get metrics for a single ELB named `myLB`:
```
{
  "region": "eu-west-1",
  "delay_seconds": 600,
  "metrics": [
    {
      "aws_namespace": "AWS/ELB",
      "aws_metric_name": "RequestCount",
      "aws_dimensions": [
        "AvailabilityZone",
        "LoadBalancerName:myLB"
      ],
      "aws_statistics": ["Sum"],
      "period_seconds": 300
    }
  ]
}
```

PR #8, opened 2015-03-31
Add wildcard matching to dimension selection

fstehle:
Enhancement to PR https://github.com/prometheus/cloudwatch_exporter/pull/6
Add wildcard matching to dimension selection.

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -21,6 +21,7 @@ import java.util.Set;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
@@ -41,6 +42,7 @@ public class CloudWatchCollector extends Collector {
       List<String> awsStatistics;
       List<String> awsDimensions;
       Map<String,List<String>> awsDimensionSelect;
+      Map<String,List<String>> awsDimensionSelectRegex;
       String help;
     }
 
@@ -104,8 +106,14 @@ public class CloudWatchCollector extends Collector {
           if (jsonMetricRule.containsKey("aws_dimensions")) {
             rule.awsDimensions = (JSONArray)jsonMetricRule.get("aws_dimensions");
           }
+          if (jsonMetricRule.containsKey("aws_dimension_select") && jsonMetricRule.containsKey("aws_dimension_select_regex")) {
+            throw new IllegalArgumentException("Must not provide aws_dimension_select and aws_dimension_select_regex at the same time");
+          }
           if (jsonMetricRule.containsKey("aws_dimension_select")) {
             rule.awsDimensionSelect = (JSONObject)jsonMetricRule.get("aws_dimension_select");
+          }
+          if (jsonMetricRule.containsKey("aws_dimension_select_regex")) {
+            rule.awsDimensionSelectRegex = (JSONObject)jsonMetricRule.get("aws_dimension_select_regex");
           }
           if (jsonMetricRule.containsKey("aws_statistics")) {
             rule.awsStatistics = (JSONArray)jsonMetricRule.get("aws_statistics");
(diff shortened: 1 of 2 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -180,7 +183,7 @@ private boolean useMetric(MetricRule rule, Metric metric) {
         String dimensionValue = dimension.getValue();
         if (dimensionSelectKeys.contains(dimensionName)) {
           List<String> allowedDimensionValues = rule.awsDimensionSelect.get(dimensionName);
-          if (!allowedDimensionValues.contains(dimensionValue)) {
```
brian-brazil: I think this needs a separate setting, as there's cases where we'd want to pass the aws_dimension_select directly to AWS (the billing API docs say it needs this for example).

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -188,6 +191,38 @@ private boolean useMetric(MetricRule rule, Metric metric) {
...
+          return true;
+        }
+      }
+      return false;
+    }
+
+    /**
+     * Check if a string with wildcard characters matches a given input value
+     */
+    protected static boolean wildcardMatch(String pattern, String input) {
```
brian-brazil: Let's just go with regexes

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
fstehle: I do not understand the use case. 
Can you explain this in more detail please?

brian-brazil: http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/billing-metricscollected.html See the currency dimension.

fstehle:
So I would suggest having an additional property `aws_dimension_select_regex`, eg.
```
"aws_dimension_select_regex": {
  "LoadBalancerName": ["prod-(.*)"]
}
```
If `aws_dimension_select_regex` and `aws_dimension_select` are set at the same time all dimensions which match the regex in `aws_dimension_select_regex` or the string in `aws_dimension_select` are returned.

fstehle:
@brian-brazil Could you have a look on the updated commit please?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -168,12 +173,22 @@ private CloudWatchCollector(JSONObject config, AmazonCloudWatchClient client) th
     }
 
     /**
-     * Check if a metric should be used according to `aws_dimension_select`
+     * Check if a metric should be used according to `aws_dimension_select` or `aws_dimension_select_regex`
      */
     private boolean useMetric(MetricRule rule, Metric metric) {
-      if (rule.awsDimensionSelect == null) {
+      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null) {
```
brian-brazil: Will this cause an NPE is only one is null?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -168,12 +173,22 @@ private CloudWatchCollector(JSONObject config, AmazonCloudWatchClient client) th
...
+        return true;
+      }
+      if (metricsIsInAwsDimensionSelect(rule, metric) || metricIsInAwsDimensionSelectRegex(rule, metric)) {
```
brian-brazil: This should be an and I'd say (not that it'd make much sense - maybe forbid that as a config error?)

brian-brazil:
Can you add unittests please?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
fstehle: Why should we disallow that users can specify some AWS metrics with an exact string match and some metrics with an regex match? 

For example the following configuration to get all load balancers metrics with the name `myelb1`, `myelb2` or the prefix `anotherelb`:
```
{ 
  "region": "eu-west-1",
  "delay_seconds": 600,
  "metrics": [
    {
      "aws_namespace": "AWS/ELB", "aws_metric_name": "HealthyHostCount",
      "aws_dimensions": ["LoadBalancerName"], "aws_statistics": ["Minimum"],
      "aws_dimension_select": {
        "LoadBalancerName": ["myelb1", "myelb2"]
      },
      "aws_dimension_select_regex": {
        "LoadBalancerName": ["anotherelb(.*)"]
      },
      "period_seconds": 300
    }
  ]
}
```

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
fstehle: You are right. I will fix that.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
brian-brazil: I could see that be confusing - is that an AND or and OR? Given how the rest works I'd expect it to be an AND.

fstehle: I changed this.
Now it is forbidden to  set `aws_dimension_select` and `aws_dimension_select_regex` at the same time:
https://github.com/Jimdo/cloudwatch_exporter/blob/dimension_value_wildcard/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java#L110

fstehle:
I added a unittest for it.

fstehle:
@brian-brazil Could you please have another look at this PR?

brian-brazil:
Thanks!

PR #12, opened 2015-10-02
Use the help text when it is specified in the rule

r4k:
Use the help text if it is provided in the description, otherwise use the generated help text.

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -259,9 +259,9 @@ public class CloudWatchCollector extends Collector {
     }
 
     private String help(MetricRule rule, String unit, String statistic) {
-        if (rule.help != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -259,6 +259,9 @@ private String safeName(String s) {
     }
 
     private String help(MetricRule rule, String unit, String statistic) {
+        if (rule.help != null) {
+            return rule.help;
+        }
```
grobie: Indentation seems to be off by two spaces compared to the rest of the file.

beorn7:
@brian-brazil

brian-brazil:
This looks good, please address @grobie's comment and if you could remove beorn's email from the commit that'd be appreciated.

r4k:
Fixing line spacing

brian-brazil:
Thanks!

PR #13, opened 2015-10-02
@beorn@soundcloud.com -- Use the help text when it is specified in th…

r4k:
…e rule

Using the help field when it is specified in the definition. Otherwise the generated help text is returned like earlier. Also, Fixing the spaces.


beorn7:
Duplicate of https://github.com/prometheus/cloudwatch_exporter/pull/12 .

PR #14, opened 2015-10-15
Added YAML support, updated example config

inondle:
Hi, 

Brian said this project also needed conversion from JSON to YAML. I used snakeYaml to update the CloudwatchCollector and it's corresponding test cases. I also updated the example configuration file that is given. Let me know if anything needs improvement. 

Regards, 
Inondle


example.yml
```diff
@@ -0,0 +1,188 @@
+---
+region: eu-west-1
+metrics:
+- aws_namespace: AWS/ELB
+  aws_metric_name: HealthyHostCount
+  aws_dimensions:
```
brian-brazil: Hmm, I think the length of this file might be a little scary to users, would you mind doing `aws_dimensions: [AvailabilityZone, LoadBalancerName]` and the same for statistics?

inondle:
Updated the list notation for the example config file and added some whitespace so it doesn't look so monotonous.

brian-brazil:
Great, thanks!

Issue #23, opened 2016-02-22
Outdated dependency in pom.xml


JavierZunzunegui:
snakeyaml 1.17 appears to no longer be available, on 1.18 now.
https://oss.sonatype.org/content/repositories/snapshots/org/yaml/snakeyaml/

brian-brazil:
Thanks for the report, we shouldn't be using SNAPSHOT versions at all. Would you like to send a PR?

JavierZunzunegui:
Made PR https://github.com/prometheus/cloudwatch_exporter/pull/24 using non-SNAPSHOT snakeyaml.

PR #24, opened 2016-02-22
org.yaml.snakeyaml 1.17 without SNAPSHOT

JavierZunzunegui:
As per https://github.com/prometheus/cloudwatch_exporter/issues/23


brian-brazil:
Thanks!

PR #27, opened 2016-03-24
introduce optional role_arn for retrieving cross account metrics.

schnipseljagd:
This allows us to start a cloudwatch_exporter which can query metrics from another AWS account.

cc: @bracki

The cloudformation template we use:
```json
{
    "Resources": {
        "GrantReadAccessToAnotherAccount": {
            "Properties": {
                "AssumeRolePolicyDocument": {
                    "Statement": [
                        {
                            "Action": [
                                "sts:AssumeRole"
                            ],
                            "Effect": "Allow",
                            "Principal": {
                                "AWS": [
                                    "arn:aws:iam::YOUR_ACCOUNT_HERE:root"
                                ]
                            }
                        }
                    ],
                    "Version": "2012-10-17"
                },
                "ManagedPolicyArns": [
                    "arn:aws:iam::aws:policy/CloudWatchReadOnlyAccess"
                ]
            },
            "Type": "AWS::IAM::Role"
        }
    }
}

```

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -1,5 +1,6 @@
 package io.prometheus.cloudwatch;
 
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
 import com.amazonaws.services.cloudwatch.model.Datapoint;
 import com.amazonaws.services.cloudwatch.model.Dimension;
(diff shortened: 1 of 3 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -83,7 +84,15 @@ private CloudWatchCollector(Map<String, Object> config, AmazonCloudWatchClient c
         }
 
         if (client == null) {
-          this.client = new AmazonCloudWatchClient();
+          if (config.containsKey("role_arn")) {
+            STSAssumeRoleSessionCredentialsProvider credentialsProvider = new STSAssumeRoleSessionCredentialsProvider(
```
brian-brazil: Does this handle refreshing credentials as needed, or does that need to be implemented by us?

bracki: That is handled by the credentials provider as can be seen [here](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.java#L271).

schnipseljagd: I just waited an hour and exporter still worked :-)

brian-brazil:
Thanks!

suryam27:
Does this feature work for AssumeRoleWithWebIdentity, I'm trying to setup cross account assumerole with OIDC but it doesn't seem to work.

PR #30, opened 2016-04-28
Support regex in config

harmw:
Our use-case requires stateless containers, not backed using static configuration files. Using the environment to specify on which values to match, the operator simply deploys this **one** container to each of our Mesos deployments with some env vars.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,8 @@ RUN echo 'http://dl-3.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositor
     && apk add maven openjdk8-jre \
     && mvn package \
     && mv target/cloudwatch_exporter-*-with-dependencies.jar /cloudwatch_exporter.jar \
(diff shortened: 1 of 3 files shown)


harmw:
@brian-brazil Please take a look :smiley:

brian-brazil:
Thanks for the PR, we usually leave this sort of thing up to the configuration management rather than trying to support many different configuration approaches in the binary itself. I'd suggest using a little bit of sed to produce the config file you need from environment variables, rather than adding this to the exporter..

harmw:
I explicitly didn't want that, because it means creating some shell wrapper that would apply a bunch of `sed` tricks on the config.yml before starting the main `java` process. This approach seems a lot cleaner and we're using it in our Scala based projects as well.

brian-brazil:
Our principle with Prometheus is that we keep our config completely dumb, and leave any templating to the user's existing configuration management. We have to consider overall complexity and cleanliness of the project, even if it means some users have to do a small bit more work.

PR #39, opened 2016-09-16
Add healthcheck to /status route

fstehle:
in order to have a healthcheck endpoint which could be used by a load balancer (eg. AWS ELB).

This avoids that one has to use the /metrics route for a healthcheck, which is very expensive and slow when a lot of metrics are configured to be collected.

PTAL @brian-brazil

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/HealthCheckServlet.java
+++ b/src/main/java/io/prometheus/cloudwatch/HealthCheckServlet.java
@@ -1,0 +1,38 @@
+package io.prometheus.cloudwatch;
+
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+public class HealthCheckServlet extends HttpServlet {
+    private AmazonCloudWatchClient client;
+
+    public HealthCheckServlet(AmazonCloudWatchClient client) {
+        this.client = client;
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/io/prometheus/cloudwatch/HealthCheckServlet.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public HealthCheckServlet(AmazonCloudWatchClient client) {
+        this.client = client;
+    }
+
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        PrintWriter writer = resp.getWriter();
+        resp.setContentType("text/plain; charset=utf-8");
+        try {
+            this.client.listMetrics();
```
brian-brazil: This costs money.

brian-brazil:
Exporters should not be put behind any load balancers, Prometheus should talk to them directly.

fstehle:
Yes, however in our setup we need an endpoint with  a HTTP status code 200 reponse. 
What would you suggest to do?
Changing the HTTP status code of the / route?

brian-brazil:
I'd suggest changing your setup so the cloudwatch exporter runs beside Prometheus.
We don't generally accept changes to allow for non-recommended deployment models.

fstehle:
Yes, I totally see your point. But why not have a simple home page under the / route like the `node_exporter` in https://github.com/prometheus/node_exporter/blob/60cbc9efc0ff3a5c25394aeaef0c8df3ebad7037/node_exporter.go#L166:L174 (introduced in https://github.com/prometheus/node_exporter/pull/32)?

brian-brazil:
I'd accept a change to add a simple home page.

fstehle:
I have created a separate PR for this: https://github.com/prometheus/cloudwatch_exporter/pull/40

PR #43, opened 2016-11-07
Update pom.xml

shidel-dev:
update aws sdk to 	1.11.52
https://aws.amazon.com/releasenotes/Java/9979983567247718
does not look like this is using any of the breaking changes described here.

The provided docker image can not be run with ECS without this update due to the way that ECS tasks get their credentials.
Built container locally and verified that tests are passing.


brian-brazil:
Thanks!

PR #48, opened 2017-01-31
Upgrade pom.xml dependencies for proxy support

gozer:
I was trying to use this exporter behind an http proxy, and after upgrading the following libraries, I was successfull.

 - com.amazonaws.aws-java-sdk-cloudwatch => 1.11.86
 - com.amazonaws.aws-java-sdk-sts        => 1.11.86
 - io.prometheus.simpleclient            => 0.0.20
 - io.prometheus.simpleclient_servlet    => 0.0.20

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,12 +46,12 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-cloudwatch</artifactId>
-      <version>1.11.52</version>
+      <version>1.11.86</version>
     </dependency>
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-sts</artifactId>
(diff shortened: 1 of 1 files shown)


brian-brazil:
If we're going to that version of client_java, then the collector should implement an empty decribe() method.

gozer:
I must aplogize here, as I imagine that by client_java you are talking about io.prometheus.simpleclient?

After some more testing, I found that only upgrading
 - com.amazonaws.aws-java-sdk-cloudwatch => 1.11.86
 -  com.amazonaws.aws-java-sdk-sts => 1.11.86

Also makes the http proxy issue go away.

brian-brazil:
You can update the PR to that then

gozer:
PR updated!

brian-brazil:
Thanks!

PR #52, opened 2017-02-26
configurable metrics path

moshebs:
It is very useful to be able to control the metrics path, currently hardcoded. Very small change :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.iml
+.idea/*
+target/*
--- a/pom.xml
+++ b/pom.xml
@@ -43,11 +43,16 @@
 
   <dependencies>
 
-    <dependency>
-      <groupId>com.amazonaws</groupId>
-      <artifactId>aws-java-sdk-cloudwatch</artifactId>
-      <version>1.11.86</version>
-    </dependency>
+      <dependency>
+          <groupId>com.amazonaws</groupId>
+          <artifactId>aws-java-sdk-cloudwatch</artifactId>
+          <version>1.11.86</version>
+      </dependency>
+      <dependency>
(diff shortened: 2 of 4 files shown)


grobie:
The [common way in other Prometheus exporters is to provide a flag](https://github.com/search?q=web.telemetry-path+user%3Aprometheus&ref=searchresults&type=Code&utf8=%E2%9C%93) `--web.telemetry-path` which defaults to `/metrics` instead of using a positional argument.

moshebs:
@grobie, the port is already configured as a commandline parameter, so it seems like this exporter is designed a bit differently to begin with.

brian-brazil:
The convention is /metrics, why doesn't that work for you?

I'm hesitant to add this to this exporter as there's not really a sane way to specify it.

moshebs:
Here is why I find this option useful: I deploy Prometheus and a set of exporters in AWS and put the behind an Application Load Balancer (ALB), which routes the requests bases on path. The ALB is unable to modify the path it sends to the backend, so if all the exporters are responding to /metrics, I have no way of routing the requests to them. Could be useful in other cases as well.

brian-brazil:
The recommended way to run Prometheus is to have it run on the same network as what it's monitoring and having it talk to exporters directly.

Putting a load balancer in the way is going to be unworkable in general, particularly when you've more than one replica of a job.

moshebs:
I am not using the ALB as a load balancer per-se, actually just as a means to do service discovery. Prometheus is deployed in the same network as the monitored targets and the exporters. Think of it more as a reverse proxy.

brian-brazil:
I'd recommend you use the EC2 service discovery that comes with Prometheus.

PR #55, opened 2017-03-23
Set instance label to empty

brian-brazil:
It's expected that the user will set honor_labels
for the cloudwatch exporter, so we should
explicitly set both job and instance.


juliusv:
:+1:

Issue #18, opened 2016-01-04
Multiple same HELP/TYPE lines for DynamoDB metrics


marceldegraaf:
We're pulling DynamoDB metrics from CloudWatch, and we're interested in individual table metrics (e.g. `ConsumedReadCapacity` for a table `foo_production`), and the global secondary index metrics on a table (e.g. `ConsumedReadCapacity` for global secondary index `bar` on the `foobar_production` table).

Our configuration for `cloudwatch-exporter` looks like this:

```
- aws_namespace: "AWS/DynamoDB"
  aws_metric_name: "ProvisionedReadCapacityUnits"
  aws_dimensions:
    - TableName
  aws_dimension_select_regex:
    TableName:
      - "(.*)_production"

- aws_namespace: "AWS/DynamoDB"
  aws_metric_name: "ProvisionedReadCapacityUnits"
  aws_dimensions:
    - TableName
    - GlobalSecondaryIndexName
  aws_dimension_select_regex:
    TableName:
      - "(.*)_production"
```

However, this leads to two occurrences of the same `HELP` and `TYPE` lines for these metrics:

```
# HELP aws_dynamodb_provisioned_read_capacity_units_sum CloudWatch metric AWS/DynamoDB ProvisionedReadCapacityUnits Dimensions: [TableName] Statistic: Sum Unit: Count
# TYPE aws_dynamodb_provisioned_read_capacity_units_sum gauge
...
# HELP aws_dynamodb_provisioned_read_capacity_units_sum CloudWatch metric AWS/DynamoDB ProvisionedReadCapacityUnits Dimensions: [TableName, GlobalSecondaryIndexName] Statistic: Sum Unit: Count
# TYPE aws_dynamodb_provisioned_read_capacity_units_sum gauge
```

Prometheus doesn't like this: `text format parsing error in line 111: second HELP line for metric name "aws_dynamodb_provisioned_read_capacity_units_sum"`

I was hoping #11 would fix this, but of course even with a custom `HELP` text, the `TYPE` will still be duplicated, and thus be invalid.

Is there a way to get around this, and have both table and global secondary index data in Prometheus?

brian-brazil:
The problem here is that you're exposing two different metrics with the same name, which is strongly discouraged. I'd recommend only exporting the more granular of the two, and aggregating on the Prometheus side.

marceldegraaf:
[...]

PR #56, opened 2017-03-27
Avoid generating duplicate help text for DynamoDB

cosmopetrich:
This is my attempt at fixing the issue described in #18. I'm not too experienced with Java, so please let me know if anything looks stupid and I'll fix it up.

As was suggested in the linked issue this modifies `baseName` when scraping metrics from AWS/DynamoDB and fanning out across the GlobalSecondaryIndexName namespace.

That is, given a config.yaml like the below

```yaml
---
region: us-east-1
metrics:
 - aws_namespace: AWS/DynamoDB
   aws_metric_name: ConsumedReadCapacityUnits
   aws_dimensions: [TableName, GlobalSecondaryIndexName]
   aws_dimension_select:
     TableName: ["my-test-table"]
   aws_statistics: [Sum]
 - aws_namespace: AWS/DynamoDB
   aws_metric_name: ConsumedReadCapacityUnits
   aws_dimensions: [TableName]
   aws_dimension_select:
     TableName: ["my-test-table"]
   aws_statistics: [Sum]
```

a scrape will produce the following metrics

```
# HELP aws_dynamodb_consumed_read_capacity_units_index_sum CloudWatch metric AWS/DynamoDB ConsumedReadCapacityUnits Dimensions: [TableName, GlobalSecondaryIndexName] Statistic: Sum Unit: Count
# TYPE aws_dynamodb_consumed_read_capacity_units_index_sum gauge
aws_dynamodb_consumed_read_capacity_units_index_sum{job="aws_dynamodb",instance="",table_name="my-test-table",global_secondary_index_name="my-test-index",} 123.4
# HELP aws_dynamodb_consumed_read_capacity_units_sum CloudWatch metric AWS/DynamoDB ConsumedReadCapacityUnits Dimensions: [TableName] Statistic: Sum Unit: Count
# TYPE aws_dynamodb_consumed_read_capacity_units_sum gauge
aws_dynamodb_consumed_read_capacity_units_sum{job="aws_dynamodb",instance="",table_name="my-test-table",} 456.7
```

I'm not 100% certain about the metric name, `aws_dynamodb_METRIC_index_STATISTIC` vs `aws_dynamodb_METRIC_STATISTIC`. Perhaps `aws_dynamodb_index_METRIC_STATISTIC` would be better, or something else entirely?

If anyone is currently sidestepping this issue by running a separate instance of the exporter for index metrics then they'll have the metric names changed under them if they upgrade. We might want to call that out if we decide to merge this. 

Finally, I'm aware of one other case similar to DynamoDB. The [Kinesis Client Library](https://github.com/awslabs/amazon-kinesis-client) will [publish Cloudwatch metrics](http://docs.aws.amazon.com/streams/latest/dev/monitoring-with-kcl.html) with useful names like `Time` and `Success` with what I believe are different meanings in different sets of dimensions. Since this tool isn't ideal for extracting KCL metrics (each KCL application will create its own _namespace_) we can probably just ignore it for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,17 +70,9 @@ In addition `cloudwatch_exporter_scrape_error` will be non-zero if an error
 occurred during the scrape, and `cloudwatch_exporter_scrape_duration_seconds`
 contains the duration of that scrape.
 
-### Metrics for DynamoDB Global Secondary Indexes
+### Special handling for certain DynamoDB metrics
 
-CloudWatch exposes some DynamoDB metrics with the same names in both the
-TableName and TableName + GlobalSecondaryIndexName dimensions. These metrics
-require special handling to avoid exporting duplicate HELP and and TYPE lines
-in the case where the exporter is configured to retrieve metrics for both tables
-and indexes.
-
-When exporting one of the problematic metrics listed below for an index the
-exporter will use a metric name in the format `aws_dynamodb_METRIC_index_STATISTIC`
-rather than the usual `aws_dynamodb_METRIC_STATISTIC`.
+The DynamoDB metrics listed below break the usual CloudWatch data model.
 
  * ConsumedReadCapacityUnits
  * ConsumedWriteCapacityUnits
@@ -88,6 +80,19 @@ rather than the usual `aws_dynamodb_METRIC_STATISTIC`.
  * ProvisionedWriteCapacityUnits
  * ReadThrottleEvents
  * WriteThrottleEvents
+
+When these metrics are requested in the TableName dimension CloudWatch will
+return data only for the table itself, not for its Global Secondary Indexes.
+Retrieving data for indexes requires requesting data across both the TableName
+and GlobalSecondaryIndexName dimensions. This behaviour is different to that
+of every other CloudWatch namespace and requires that the exporter handle these
+metrics differently to avoid generating duplicate HELP and TYPE lines.
+
+When exporting one of the problematic metrics for an index the exporter will use
+a metric name in the format `aws_dynamodb_METRIC_index_STATISTIC` rather than
+the usual `aws_dynamodb_METRIC_STATISTIC`. The regular naming scheme will still
+be used when exporting these metrics for a table, and when exporting any other
+DynamoDB metrics not listed above.
 
 ### Cost
 
--- a/README.md
+++ b/README.md
@@ -70,6 +70,25 @@ In addition `cloudwatch_exporter_scrape_error` will be non-zero if an error
 occurred during the scrape, and `cloudwatch_exporter_scrape_duration_seconds`
 contains the duration of that scrape.
 
+### Metrics for DynamoDB Global Secondary Indexes
+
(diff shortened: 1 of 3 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -309,6 +309,15 @@ private void scrape(List<MetricFamilySamples> mfs) {
 
         String unit = null;
 
+        if (rule.awsNamespace.equals("AWS/DynamoDB") && rule.awsDimensions.contains("GlobalSecondaryIndexName")) {
```
brian-brazil: Are there any metrics exposed with this label that don't have this flaw?

cosmopetrich: The metrics that can currently result in duplicate help text are the following:

 * ConsumedReadCapacityUnits
 * ConsumedWriteCapacityUnits
 * ProvisionedReadCapacityUnits
 * ProvisionedWriteCapacityUnits
 * ReadThrottleEvents
 * WriteThrottleEvents

There are also some metrics that should only ever appear in [TableName, GlobalSecondaryIndexName], never in just [TableName].

 * OnlineIndexConsumedWriteCapacity
 * OnlineIndexPercentageProgress
 * OnlineIndexThrottleEvents

My thinking was that it was desirable to have the suffix on all index-related metrics. Would it be better to attach it only for the ones that are potentially duplicates?

brian-brazil: We should only attach it on the broken metrics. This will need a quick mention in the readme too.

cosmopetrich:
@brian-brazil I've pushed a change that attaches the suffix only to the problematic metrics and updated the readme. It _appears_ to work in real-world testing as well, though testing the OnlineIndex* metrics requires creating a new index since those metrics will be absent from CloudWatch once the index comes online.

README.md
```diff
@@ -69,6 +69,25 @@ In addition `cloudwatch_exporter_scrape_error` will be non-zero if an error
 occurred during the scrape, and `cloudwatch_exporter_scrape_duration_seconds`
 contains the duration of that scrape.
 
+### Metrics for DynamoDB Global Secondary Indexes
+
+CloudWatch exposes some DynamoDB metrics with the same names in both the
+TableName and TableName + GlobalSecondaryIndexName dimensions. These metrics
```
brian-brazil: I'd make it more explicit that these break the usual cloudwatch data model.

brian-brazil:
You have conflicts that need resolving.

cosmopetrich:
> You have conflicts that need resolving.

Sorry about that. I was planning to wait until we'd sorted everything out before rebasing, but I've gone ahead and done it now.

I've taken another stab at the readme as well. Let me know if it's any better.

brian-brazil:
Thanks!

PR #61, opened 2017-04-20
Add ELB tags to metrics

TomaszSekTesco:
It would be nice to have ELB tags in metrics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ This includes the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment
 variables.
 
 The `cloudwatch:ListMetrics` and `cloudwatch:GetMetricStatistics` IAM permissions are required.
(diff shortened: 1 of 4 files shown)


brian-brazil:
Thanks for the PR, this has already been discussed in several issues. Adding this for every possible attribute for every possible AWS entity is not sane.

antoniaklja:
@brian-brazil I don't agree. Let assume situation when you have Prometheus+Alertmanager and generic, meanigless ELB metrics attributes. How will you recognize specific ELB for instance in kubernetes cluster?
Lucky guess? don't think so..

brian-brazil:
I don't see what that has to do with my objection.

I'd suggest you look into having more meaningful ELB names.

antoniaklja:
So I'm forced to write my own better implementation. cya

2017-04-25 11:33 GMT+02:00 Brian Brazil <notifications@github.com>:

PR #78, opened 2017-10-10
Configuration reload endpoint

alkalinecoffee:
Adds an endpoint to dynamically reload the yaml configuration via `curl -X POST localhost:9106/-/reload`.

The `rules` list is now wrapped in a `synchronized` block for consistent object access during updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,9 +99,9 @@ DynamoDB metrics not listed above.
 There are two ways to reload configuration:
 
 1. Send a SIGHUP signal to the pid: `kill -HUP 1234`
-2. POST to the `reload` endpoint: `curl -X localhost:9107/-/reload`
+2. POST to the `reload` endpoint: `curl -X localhost:9106/-/reload`
 
-If an error occurs during the reload, see check the exporter's log output.
+If an error occurs during the reload, check the exporter's log output.
 
 ### Cost
 
--- a/README.md
+++ b/README.md
@@ -94,6 +94,15 @@ the usual `aws_dynamodb_METRIC_STATISTIC`. The regular naming scheme will still
 be used when exporting these metrics for a table, and when exporting any other
 DynamoDB metrics not listed above.
 
+### Reloading Configuration
+
+There are two ways to reload configuration:
+
+1. Send a SIGHUP signal to the pid: `kill -HUP 1234`
+2. POST to the `reload` endpoint: `curl -X localhost:9107/-/reload`
+
+If an error occurs during the reload, see check the exporter's log output.
+
 ### Cost
 
 Amazon charges for every API request, see the [current charges](http://aws.amazon.com/cloudwatch/pricing/).
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -14,6 +14,8 @@ import com.amazonaws.services.cloudwatch.model.ListMetricsResult;
 import com.amazonaws.services.cloudwatch.model.Metric;
 import io.prometheus.client.Collector;
 import io.prometheus.client.Counter;
+
+import java.io.FileReader;
 import java.io.Reader;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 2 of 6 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -160,9 +184,11 @@ private CloudWatchCollector(Map<String, Object> config, AmazonCloudWatchClient c
             rule.delaySeconds = defaultDelay;
           }
         }
+
+        setRules(rules);
```
brian-brazil: The rules and client should be set atomically, in case someone changes both.

src/main/java/io/prometheus/cloudwatch/DynamicReloadServlet.java
```diff
@@ -0,0 +1,32 @@
...
+    private static String configFilePath;
+
+    private static final Logger LOGGER = Logger.getLogger(CloudWatchCollector.class.getName());
+
+    public DynamicReloadServlet(CloudWatchCollector cc, String path) {
+        this.collector = cc;
+        this.configFilePath = path;
+    }
+
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
```
brian-brazil: Can you add a doGet with a message telling the user to use POST? We found that cut down support requests a lot for Prometheus.

src/main/java/io/prometheus/cloudwatch/DynamicReloadServlet.java
```diff
@@ -0,0 +1,32 @@
...
+        LOGGER.log(Level.INFO, "Reloading configuration file");
+
+        this.collector.loadConfig(new FileReader(this.configFilePath));
+
+        resp.setContentType("text/html");
```
brian-brazil: This is text/plain

src/test/java/io/prometheus/cloudwatch/CloudWatchCollectorTest.java
```diff
@@ -297,15 +297,6 @@ public void testGetDimensionsUsesNextToken() throws Exception {
   }
 
   @Test
-  public void testGetMonitoringEndpoint() throws Exception {
```
brian-brazil: Why remove this test?

alkalinecoffee:
Made updates per feedback; please review again

src/test/java/io/prometheus/cloudwatch/CloudWatchCollectorTest.java
alkalinecoffee: I initially removed it because it's testing internal implementation, but also realized it's not being tested elsewhere in the test suite.  Re-added in 13a857c.

src/main/java/io/prometheus/cloudwatch/WebServer.java
```diff
@@ -2,27 +2,41 @@
...
+    public static void main(String[] args) throws Exception {
+        if (args.length < 2) {
+            System.err.println("Usage: WebServer <port> <yml configuration file>");
+            System.exit(1);
+        }
+
+        configFilePath = args[1];
+        collector = new CloudWatchCollector(new FileReader(configFilePath)).register();
+
+        ReloadSignalHandler.start();
```
brian-brazil: I'd inline this rather than using public static variables.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -74,25 +79,27 @@ private CloudWatchCollector(Map<String, Object> config, AmazonCloudWatchClient c
         loadConfig(config, client);
     }
 
-    private ArrayList<MetricRule> getRules() {
-        synchronized (rules) {
-            return rules;
+    private AmazonCloudWatchClient getClient() {
```
brian-brazil: This is still racey. Both need to be fetched at the same time for a given scrape.

alkalinecoffee: Made some changes in 9705710 to clone the active config that's in memory.

src/main/java/io/prometheus/cloudwatch/WebServer.java
alkalinecoffee: We can do that, but we'd just be moving the `collector` var from `WebServer` to somewhere else (like `ReloadSignalHandler`) since we need to reference it from both the signal handler and `DynamicReloadServlet`.

Where are you thinking we should store this value, then?

README.md
```diff
@@ -94,6 +94,15 @@ the usual `aws_dynamodb_METRIC_STATISTIC`. The regular naming scheme will still
...
 DynamoDB metrics not listed above.
 
+### Reloading Configuration
+
+There are two ways to reload configuration:
+
+1. Send a SIGHUP signal to the pid: `kill -HUP 1234`
+2. POST to the `reload` endpoint: `curl -X localhost:9107/-/reload`
+
+If an error occurs during the reload, see check the exporter's log output.
```
brian-brazil: s/see //

src/main/java/io/prometheus/cloudwatch/WebServer.java
brian-brazil: I'd pass the collector into the handler and servlet.

alkalinecoffee: Gotcha--I misread what you said.  Changed in a014d06.

brian-brazil:
Thanks!

alkalinecoffee:
🍻

PR #85, opened 2018-01-15
Add lambda metrics

HarryKroes:


Tilian:
Uhh... I don't think this is the repository we're trying to merge to.

brian-brazil:
Nope, but expanding config.yml with Lamda examples would be appreciated.

PR #87, opened 2018-02-12
Generate default data for metrics when cloudwatch returns nothing.

rsgoodall:
To return data (values default to 0) when it is missing from cloudwatch. 
For example with ELBs  data is often missing if there hasn't been any 4XX etc requests.


brian-brazil:
What exactly are you proposing to do?

We merely mirror the data coming from Cloudwatch, we should not be making it up.

PR #90, opened 2018-03-01
Retry throttled CloudWatch queries

kppullin:
This PR intends add support for retrying CloudWatch queries when they are throttled.

The code will retry queries 10 times. The backoff duration is applied via a simple exponential backoff algorithm with a max value of 10 seconds.

There is also a new counter, `cloudwatchRateExceeded`, which exposes the number of times the a request was throttled.  One use case here is to alert on high rates of throttling, so that one knows to request an increase on the API limits.

It may be useful to expose the retry attempt limit, max sleep limit, and base sleep limit via the config file.  If others see a need I can make that change.

/cc @brian-brazil per contributing guidelines

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -4,6 +4,7 @@ import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
+import com.amazonaws.services.cloudwatch.model.AmazonCloudWatchException;
 import com.amazonaws.services.cloudwatch.model.Datapoint;
 import com.amazonaws.services.cloudwatch.model.Dimension;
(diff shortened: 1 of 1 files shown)


brian-brazil:
Blind retries are generally a bad idea as they can lead to thundering herds, an if you're already hitting a throttle more queries doesn't help. See https://www.robustperception.io/its-overloaded-try-harder/

If you're hitting this you either need to reduce the number of queries you're making, or get your quota increased.

kppullin:
Yeah, good point - I can see retries causing problems in the general case and worth avoiding.  Another potential issue that popped in my head occurs when the retry rates are high, but eventually succeed instead of blowing up via the exception.  If you're collecting lots of metrics, combined with multiple long delays, it prevents the timely collection & reporting of metrics.

I think what would be most useful for me at this stage, is just knowing that we're being throttled so that we can request an increase of our CloudWatch limits.  Would you accept a patch that just includes the counter that tracks occurences of throttle exceptions?  Our tooling doesn't work well with triggering alerts off log messages.

brian-brazil:
> Would you accept a patch that just includes the counter that tracks occurences of throttle exceptions?

Sure.

kppullin:
Wow, that was a fast response : )  I'll work on a new PR for that.  Thanks!

PR #91, opened 2018-03-01
Expose internal CloudWatch error metrics

kppullin:
Added a new counter for tracking CloudWatch API client exceptions.

The counter includes labels for:
- The AWS provided `error code`
- The name of the CloudWatch API method that failed

Failures are _only_ recorded. Exceptions are rethrown and continue to bubble up the stack.

/cc @brian-brazil

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -4,6 +4,7 @@ import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
+import com.amazonaws.services.cloudwatch.model.AmazonCloudWatchException;
 import com.amazonaws.services.cloudwatch.model.Datapoint;
 import com.amazonaws.services.cloudwatch.model.Dimension;
 import com.amazonaws.services.cloudwatch.model.DimensionFilter;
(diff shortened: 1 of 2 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -216,8 +221,15 @@ public String getMonitoringEndpoint(Region region) {
...
       do {
         request.setNextToken(nextToken);
-        ListMetricsResult result = client.listMetrics(request);
-        cloudwatchRequests.inc();
+        ListMetricsResult result;
+        try {
+          result = client.listMetrics(request);
+          cloudwatchRequests.inc();
+        } catch (AmazonCloudWatchException ex) {
+          cloudwatchErrorRequests.labels(ex.getErrorCode(), "listMetrics").inc();
```
brian-brazil: How many values can the error code have?

kppullin: It's hard to definitively say as it's not well documented. Based on looking at error codes for a variety of AWS services, the possible values are limited do not include "unique" data that'd cause the cardinality to balloon (unique data, such as an instance id, appears in the 'message' field).

The most codes I've run across is for EC2 (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html), if that gives any sense of worst-case distinct values for the error code.

The error values themselves are nice to have, but not a necessity.  It just means that you'd have to take an extra step and search the `cloudwatch_exporter` logs to determine the underlying cause.

brian-brazil: That's a lot of error codes in the worst case. How many have you seen in practice?

You'll already know if the scrape failed due to `up` being 0.

kppullin: I've seen just a handful of error codes in practice.

AWS is pushing back on increasing the CloudWatch limits, so I'm going to need to see if I can determine what all else is querying CloudWatch on my side and possibly tune that down.

In the meantime I can monitor `cloudwatch_requests_total` to see if it stops increasing to know if we're getting heavily rate limited.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -63,6 +64,10 @@ public Object clone() throws CloneNotSupportedException {
     private static final Counter cloudwatchRequests = Counter.build()
       .name("cloudwatch_requests_total").help("API requests made to CloudWatch").register();
 
+    private static final Counter cloudwatchErrorRequests = Counter.build()
+      .name("cloudwatch_error_total").help("API requests that failed")
```
brian-brazil: errors

brian-brazil:
This has gone stale, plus failed scrapes would let you detect this.

PR #102, opened 2018-06-01
Toggle Cloudwatch timestamps on a per-metric basis

### LINKED ISSUE #100: Prometheus can drop metrics due to Cloudwatch timestamps being too old
Hi @brian-brazil! Thanks for your work on this super-useful project :)

I'm trying to scrape some metrics from S3. As per #38 I ran into issues with time range due to S3 only exposing metrics once a day, so I added `range_seconds: 86400` to ensure they get picked up.

Since #99 added Cloudwatch timestamps, the metric exposed now has an old timestamp that could be anything up to 24 hours old. Consequently I get `msg="Error on ingesting samples that are too old or are too far into the future" num_dropped=1` from Prometheus when I try and scrape the cloudwatch_exporter (off topic but out of interest; what are the semantics Prometheus uses to decide if a sample is "too old" or "too far into the future"?)

Here's the config I'm using:
```
    ---
    region: eu-west-1
    metrics:

    - aws_namespace: AWS/S3
      aws_metric_name: BucketSizeBytes
      aws_dimensions: [BucketName, StorageType]
      aws_dimension_select:
        BucketName: [my-bucket]
        StorageType: [StandardStorage]
      aws_statistics: [Average]
      range_seconds: 86400
```

Ideally I think we'd be able to toggle Cloudwatch timestamps on/off a per-metric basis (perhaps with config like `include_cloudwatch_timestamp: false`). I'm happy to raise a PR if you think it's the best way forward 👍

milesbxf:
Resolves #100 by allowing users to configure whether Cloudwatch timestamps are sent with metrics

Have tested locally and in our own environment 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ aws_extended_statistics | Optional. A list of extended statistics to retrieve. E
 delay_seconds | Optional. The newest data to request. Used to avoid collecting data that has not fully converged. Defaults to 600s. Can be set globally and per metric.
 range_seconds | Optional. How far back to request data for. Useful for cases such as Billing metrics that are only set every few hours. Defaults to 600s. Can be set globally and per metric.
 period_seconds | Optional. [Period](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/cloudwatch_concepts.html#CloudWatchPeriods) to request the metric for. Only the most recent data point is used. Defaults to 60s. Can be set globally and per metric.
-cloudwatch_timestamp | Optional. Boolean for whether to set the Prometheus metric timestamp as the original Cloudwatch timestamp. For some metrics which are updated very infrequently (such as S3/BucketSize), Prometheus may refuse to scrape them if this is set to true (see #100). Defaults to true. Can be set globally and per metric.
+set_timestamp | Optional. Boolean for whether to set the Prometheus metric timestamp as the original Cloudwatch timestamp. For some metrics which are updated very infrequently (such as S3/BucketSize), Prometheus may refuse to scrape them if this is set to true (see #100). Defaults to true. Can be set globally and per metric.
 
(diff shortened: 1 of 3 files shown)


src/test/java/io/prometheus/cloudwatch/CloudWatchCollectorTest.java
```diff
@@ -1,7 +1,6 @@
 package io.prometheus.cloudwatch;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
+import static org.junit.Assert.*;
```
brian-brazil: Please leave imports explicit

README.md
```diff
@@ -49,6 +49,7 @@ aws_extended_statistics | Optional. A list of extended statistics to retrieve. E
 delay_seconds | Optional. The newest data to request. Used to avoid collecting data that has not fully converged. Defaults to 600s. Can be set globally and per metric.
 range_seconds | Optional. How far back to request data for. Useful for cases such as Billing metrics that are only set every few hours. Defaults to 600s. Can be set globally and per metric.
 period_seconds | Optional. [Period](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/cloudwatch_concepts.html#CloudWatchPeriods) to request the metric for. Only the most recent data point is used. Defaults to 60s. Can be set globally and per metric.
+cloudwatch_timestamp | Optional. Boolean for whether to set the Prometheus metric timestamp as the original Cloudwatch timestamp. For some metrics which are updated very infrequently (such as S3/BucketSize), Prometheus may refuse to scrape them if this is set to true (see #100). Defaults to true. Can be set globally and per metric.
```
brian-brazil: `set_timestamp` maybe?

brian-brazil:
Thanks!

PR #108, opened 2018-06-18
Move config file to cmd so it can be set when running the container

MartinLeedotOrg:
Signed-off-by: Martin Lee <martin@billforward.net>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,4 +12,5 @@ WORKDIR /
 RUN mkdir /config
 
 ONBUILD ADD config.yml /config/
-ENTRYPOINT [ "java", "-jar", "/cloudwatch_exporter.jar", "9106", "/config/config.yml" ]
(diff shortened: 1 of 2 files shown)


MartinLeedotOrg:
@brian-brazil

Dockerfile
```diff
@@ -12,4 +12,5 @@ WORKDIR /
 RUN mkdir /config
 
 ONBUILD ADD config.yml /config/
-ENTRYPOINT [ "java", "-jar", "/cloudwatch_exporter.jar", "9106", "/config/config.yml" ]
+ENTRYPOINT [ "java", "-jar", "/cloudwatch_exporter.jar", "9106"]
```
brian-brazil: Why move the config and not the port?

MartinLeedotOrg: Port mapping can be changed easily with `-p 91xx:9106`

Can move it in to the CMD if you'd like though.

brian-brazil: Makes sense

brian-brazil:
Thanks!

PR #110, opened 2018-06-19
Add CloudFront examples

Mornor:
Add CloudFront exampless

- Region must be set to `us-east-1`
- `aws_dimensions` must be `[DistributionId, Region]`
- `aws_dimensions_select` must be `Region: [Global]`

See: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cf-metricscollected.html

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -141,6 +141,7 @@ metrics:
   aws_dimensions: [NodeID, ClusterIdentifier]
   aws_statistics: [Average]
 
+# For CloudFront metrics, you have to set the region to us-east-1 
 - aws_namespace: AWS/CloudFront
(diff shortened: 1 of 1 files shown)


example.yml
```diff
@@ -140,3 +140,45 @@ metrics:
   aws_metric_name: WriteIOPS
   aws_dimensions: [NodeID, ClusterIdentifier]
   aws_statistics: [Average]
+
+- aws_namespace: AWS/CloudFront
```
brian-brazil: You should add a comment about what the region needs to be up top

brian-brazil:
Thanks!

Issue #32, opened 2016-05-12
Add a health check endpoint


Luzifer:
> As a user of the CloudWatch exporter I want to be able to deploy the exporter on ECS behind an ELB without burning money at huge scale

**Current situation**

- The ELB needs an health check endpoint which responds with `HTTP200`
- The only exposed path in the exporter is `/metrics`
- Every request against `/metrics` does CloudWatch calls
- Health checks (every 10s N CloudWatch requests) are expensive

**Expected situation**

- The ELB needs an health check endpoint which responds with `HTTP200`
- There is an exposed path `/status` which responds with `HTTP200` reporting everything is fine (at least the Java application is available to respond)
- Health check requests do not burn money
- (I don't need to explain the huge AWS bill to my boss 😄 )

brian-brazil:
I'd suggest going with a simpler setup where you'd colocate the cloudwatch exporter with prometheus. 

We'd generally recommend not to put any load balancers in front of exporters, as they're just a potential failure mode.

dene14:
Well, in some cases it's not true... Cloudwatch exporter is a typical microservice that people normally want to have hosted in docker.
For AWS it's ECS service and if you don't want to go too far with service discovery you should add it to ELB or ALB. If with ELB it's possible just to check if port open, with ALB (application load balancer) you have to define some handler...

So can we please have it at least for the time while this exporter is so dumb (no caching, no ability to control specific groups of metrics in request and etc)?

PR #116, opened 2018-06-27
add health/ready endpoints

karlhungus:
#### Problem

I wanted to have a k8s pod for running my cloud watch exporter, but no readiness/health endpoint exists

#### Solution

Add one. These seem a little bit silly (they both just respond if the server is running, and don't say anything about the status of the connection to AWS (I took inspiration from prometheus: https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/web/web.go#L297-L304) 

This also addresses: https://github.com/prometheus/cloudwatch_exporter/issues/32, which was closed but... @brian-brazil

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/HealthServlet.java
+++ b/src/main/java/io/prometheus/cloudwatch/HealthServlet.java
@@ -10,6 +10,8 @@ public class HealthServlet extends HttpServlet {
 
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         resp.setContentType("text/html");
-        resp.getWriter().print("OK");
+        resp.getWriter().print("<!doctype html><html lang=en>" + 
+            "<head><meta charset=utf-8><title>cloudwatch exporter</title></head>" +
+            "<body>ok</body></html>");
     }
 }
--- a/src/main/java/io/prometheus/cloudwatch/HealthServlet.java
+++ b/src/main/java/io/prometheus/cloudwatch/HealthServlet.java
@@ -1,0 +1,15 @@
+package io.prometheus.cloudwatch;
+
+import javax.servlet.ServletException;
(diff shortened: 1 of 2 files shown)


src/main/java/io/prometheus/cloudwatch/HealthServlet.java
```diff
@@ -0,0 +1,15 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class HealthServlet extends HttpServlet {
+
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        resp.setContentType("text/html");
+        resp.getWriter().print("OK");
```
brian-brazil: This isn't valid HTML.

karlhungus: Fair, i took inspiration from the prometheus health endpoint, which I don't believe is valid html either... would:

```html
<html>ok</html>
```

be sufficient?

brian-brazil: Switching to text/plain is simplest.

karlhungus: I guess to be html valid it'd have to be:
```html
<!doctype html>
<html lang=en>
<head>
<meta charset=utf-8>
<title>cloudwatch exporter</title>
</head>
<body>ok
</body>
</html>
```

karlhungus: Thanks, i'm not a smart man :)

brian-brazil:
Thanks!

PR #126, opened 2018-07-30
Added S3 example into example.yml

jwenz723:
Signed-off-by: jeff.wenzbauer <jwenz723@gmail.com>

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -141,17 +141,19 @@ metrics:
   aws_dimensions: [NodeID, ClusterIdentifier]
   aws_statistics: [Average]
 
+# S3 Storage metrics are published to Cloudwatch 1x per day with a timestamp of midnight UTC, hence period_seconds: 86400
+# Publishing does not always occur at the same time, but it will occur before the next day, hence range_seconds: 172800
 - aws_namespace: AWS/S3
   aws_metric_name: BucketSizeBytes
   aws_dimensions: [BucketName, StorageType]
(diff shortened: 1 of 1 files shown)


example.yml
```diff
@@ -141,6 +141,20 @@ metrics:
   aws_dimensions: [NodeID, ClusterIdentifier]
   aws_statistics: [Average]
 
+- aws_namespace: AWS/S3
+  aws_metric_name: BucketSizeBytes
+  aws_dimensions: [BucketName, StorageType]
+  aws_statistics: [Maximum]
+  range_seconds: 172800
```
brian-brazil: This deserves a comment

example.yml
```diff
@@ -141,6 +141,20 @@ metrics:
   aws_dimensions: [NodeID, ClusterIdentifier]
   aws_statistics: [Average]
 
+- aws_namespace: AWS/S3
+  aws_metric_name: BucketSizeBytes
+  aws_dimensions: [BucketName, StorageType]
+  aws_statistics: [Maximum]
```
brian-brazil: Only max?

jwenz723: Changed to Average only, according to S3 Documentation: https://docs.aws.amazon.com/AmazonS3/latest/dev/cloudwatch-monitoring.html

example.yml
jwenz723: Comment added

brian-brazil:
Thanks!

PR #128, opened 2018-07-31
Move EXPOSE and MAINTAINER to final build stage

PirminTapken:
so the port and Author are set correctly in the released image.

PTAL @brian-brazil


brian-brazil:
Thanks!

PR #143, opened 2018-10-11
Do not request dimensions when they are stated

steven-aerts:
When only one dimension is defined and a `aws_dimension_select` list is provided, the dimensions are known and do not have to be requested anymore.
This can result in substantial cost savings.

In our setup where we work a lot with temporary instances, we have for example 1000 times more metrics in cloudwatch than we have live instances.
By providing the cloudwatch_exporter a list of live instances via the `aws_dimension_select` property in its configuration file, we can keep the cost cloudwatch cost more under control.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,8 @@ to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
 In the case that all `aws_dimensions` are provided in the `aws_dimension_select` list, the exporter will not perform the
-above api request, which might also reduce the cost.
+above API request.  It will request all possible combination of values for those dimensions.
+This will reduce cost as the values for the dimensions do not need to be queried anymore, assuming that all possible value combinations are present in CloudWatch.
 
 If you have 100 API requests every minute, with the price of USD$10 per million
 requests (as of Aug 2018), that is around $45 per month. The
--- a/README.md
+++ b/README.md
@@ -125,8 +125,8 @@ statistics. In addition, when `aws_dimensions` is provided, the exporter needs
 to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
-In the case only one `aws_dimensions` is provided together with its `aws_dimension_select` list, the exporter will not
-do the above api request, which might also reduce the cost.
+In the case that all `aws_dimensions` are provided in the `aws_dimension_select` list, the exporter will not perform the
+above api request, which might also reduce the cost.
 
 If you have 100 API requests every minute, with the price of USD$10 per million
 requests (as of Aug 2018), that is around $45 per month. The
--- a/README.md
+++ b/README.md
@@ -125,6 +125,9 @@ statistics. In addition, when `aws_dimensions` is provided, the exporter needs
 to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
+In the case only one `aws_dimensions` is provided together with its `aws_dimension_select` list, the exporter will not
+do the above api request, which might also reduce the cost.
+
 If you have 100 API requests every minute, with the price of USD$10 per million
 requests (as of Aug 2018), that is around $45 per month. The
 `cloudwatch_requests_total` counter tracks how many requests are being made.
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -20,6 +20,7 @@ import java.io.Reader;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Set;
@@ -210,6 +211,30 @@ public class CloudWatchCollector extends Collector {
     }
 
     private List<List<Dimension>> getDimensions(MetricRule rule, AmazonCloudWatchClient client) {
+        if (
+                rule.awsDimensions != null &&
+                rule.awsDimensions.size() == 1 &&
+                rule.awsDimensionSelect != null &&
+                rule.awsDimensionSelect.size() == 1 &&
+                rule.awsDimensionSelect.containsKey(rule.awsDimensions.get(0))
+        ) {
+            // The list of dimensions is known so no need to request it from cloudwatch.
+            String name = rule.awsDimensions.get(0);
+            List<String> values = rule.awsDimensionSelect.get(name);
+            List<List<Dimension>> dimensions = new ArrayList<List<Dimension>>(values.size());
+            for (String value : values) {
+                Dimension dimension = new Dimension();
+                dimension.setName(name);
+                dimension.setValue(value);
+                dimensions.add(Collections.singletonList(dimension));
+            }
+            return dimensions;
+        } else {
+            return listDimensions(rule, client);
+        }
+    }
+
+    private List<List<Dimension>> listDimensions(MetricRule rule, AmazonCloudWatchClient client) {
       List<List<Dimension>> dimensions = new ArrayList<List<Dimension>>();
       if (rule.awsDimensions == null) {
         dimensions.add(new ArrayList<Dimension>());
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -213,25 +213,39 @@ public class CloudWatchCollector extends Collector {
     private List<List<Dimension>> getDimensions(MetricRule rule, AmazonCloudWatchClient client) {
         if (
                 rule.awsDimensions != null &&
-                rule.awsDimensions.size() == 1 &&
                 rule.awsDimensionSelect != null &&
-                rule.awsDimensionSelect.size() == 1 &&
-                rule.awsDimensionSelect.containsKey(rule.awsDimensions.get(0))
+                rule.awsDimensions.size() > 0 &&
+                rule.awsDimensions.size() == rule.awsDimensionSelect.size() &&
+                rule.awsDimensionSelect.keySet().containsAll(rule.awsDimensions)
         ) {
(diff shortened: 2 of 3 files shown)


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -4,33 +4,19 @@
...
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
-import com.amazonaws.services.cloudwatch.model.Datapoint;
-import com.amazonaws.services.cloudwatch.model.Dimension;
-import com.amazonaws.services.cloudwatch.model.DimensionFilter;
-import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsRequest;
-import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsResult;
-import com.amazonaws.services.cloudwatch.model.ListMetricsRequest;
-import com.amazonaws.services.cloudwatch.model.ListMetricsResult;
-import com.amazonaws.services.cloudwatch.model.Metric;
+import com.amazonaws.services.cloudwatch.model.*;
```
brian-brazil: Please leave the imports explicit

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -209,6 +195,29 @@ public String getMonitoringEndpoint(Region region) {
     }
 
     private List<List<Dimension>> getDimensions(MetricRule rule, AmazonCloudWatchClient client) {
+        if (
+                rule.awsDimensions != null &&
+                rule.awsDimensions.size() == 1 &&
+                rule.awsDimensionSelect != null &&
+                rule.awsDimensionSelect.containsKey(rule.awsDimensions.get(0))
```
brian-brazil: What if there's more dimensions_select than that one provided?

brian-brazil: You're not checking for that case though.

steven-aerts: I think I am checkin it here:
```java
rule.awsDimensions.size() == 1
```

Or am I missing something.

brian-brazil: You're not doing that for awsDimensionSelect

steven-aerts: Why should I?
If the list contains `awsDimensions` only one dimension and this dimension is in the `awsDimensionSelect` map, why would we care about the other dimensions in this map?

Even in the current behavior, if the `awsDimensionSelect` contains keys which are not in the `awsDimensions`, those keys are currently ignored (as the list metrics did not select them).

Maybe I am mistaken (as I do not have many examples of cloudwatch metrics with multiple dimensions to test this with) and then I do not mind adding this extra check.

brian-brazil: I'm not sure exactly what the behaviour of cloudwatch will be here, but better to be safe than sorry if the user is doing something odd.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -209,6 +210,30 @@ public String getMonitoringEndpoint(Region region) {
     }
 
     private List<List<Dimension>> getDimensions(MetricRule rule, AmazonCloudWatchClient client) {
+        if (
+                rule.awsDimensions != null &&
+                rule.awsDimensions.size() == 1 &&
+                rule.awsDimensionSelect != null &&
+                rule.awsDimensionSelect.size() == 1 &&
+                rule.awsDimensionSelect.containsKey(rule.awsDimensions.get(0))
```
nishanth-pinnapareddy: Why do we want to limit the size of `awsDimensions` -> For example, consider this scenario I want to fetch metric `SuccessfulRequestLatency`  with dimensions `TableName, Operation` and filter out the metrics based on `TableName` == `my-table`. We can't achieve this with the above functionality.  Basically, what we need here is if there is one `awsDimensionSelect ` value for a particular `awsDimension` then we are good to go with the new way else fall back into old way.

nishanth-pinnapareddy: Also, add corresponding unit test to test this functionality.

steven-aerts: @nishanth-pinnapareddy I stuck to only one dimension as that was my current use case.  But I will do a stab at it.

On the unit tests, I will cook up something.

steven-aerts:
Kind reminder.
I think that all issues are now addressed.  Let me know if there is still something I can do.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -210,6 +211,44 @@ public String getMonitoringEndpoint(Region region) {
...
 
     private List<List<Dimension>> getDimensions(MetricRule rule, AmazonCloudWatchClient client) {
+        if (
+                rule.awsDimensions != null &&
+                rule.awsDimensionSelect != null &&
+                rule.awsDimensions.size() > 0 &&
+                rule.awsDimensions.size() == rule.awsDimensionSelect.size() &&
+                rule.awsDimensionSelect.keySet().containsAll(rule.awsDimensions)
+        ) {
+            // The full list of dimensions is known so no need to request it from cloudwatch.
```
nishanth-pinnapareddy: This won't handle the below scenario:
```
awsDimensions: [A, B]
awsDimensionSelect:
       A: [afilter]
```
If you look at my previous comment I asked you to handle these kinds of scenarios.

steven-aerts: @nishanth-pinnapareddy for the scenario above, I do no see how you can do it without triggereing a `listMetrics` call.

I thought I had addresdd your comment by allowing multple dimensions.
So in my first commit I only allowed:

```
rule.awsDimensions.size() == 1 &
```

In my second commit ed33a751to address your comments, I updated this to allow any number Dimensions to cover:
```
awsDimensions: [A, B]
awsDimensionSelect:
       A: [afilter]
       B: [bfilter]
```

But all dimensions need to be specified.

steven-aerts:
@nishanth-pinnapareddy was my response to your comments satisfying?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
nishanth-pinnapareddy: We always do listmetrics call, the problem we want to solve here passing dimension filter values to this call. If there are multiple dimension filter values we can't pass. But if there is single specified value for the dimension we can do it. So for the above example, you pass the filterValue for dimension `A` and for `B` you just pass the name. The listmetrics call will return which matches to above two conditions. What I thinking of is below:

```
if awsDimesionSelect != null {
 // check if all the dimensions have less than or equal 1 filter values.
// if the above check passes, add these values to DimensionFilter Object
// otherwise do it oldway
}
```

steven-aerts: Hello @nishanth-pinnapareddy,

I know this is possible with the aws listMetrics API, but for me this is a different commit/pull request, as it is in a different place of the code.
As this PR introduces an optimization which circumvents the listMetrics call from being called at all.

This PR is for example useful if you know upfront the resources you want to fetch the metrics from.  For example in our setup  we work a lot with temporary instances, so we have 1000 times more metrics in cloudwatch than we have live instances (as the metrics stay for months after an instance was terminated).
We just cannot afford using the listMetrics API.  So we have a script which regularly queries ec2 and fills in the dimensions we are currently interested in.

Do you agree that the optimization you propose is a different optimization which deserves a separate PR?

peimanja:
any updates?

steven-aerts:
@peimanja @brian-brazil for me all comments are addressed.  And this change is ready to be merged.

We are using this fix already a few months in production and it is saving us a lot of money.

README.md
```diff
@@ -125,6 +125,9 @@ statistics. In addition, when `aws_dimensions` is provided, the exporter needs
 to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
+In the case that all `aws_dimensions` are provided in the `aws_dimension_select` list, the exporter will not perform the
+above api request, which might also reduce the cost.
```
brian-brazil: This would do a full cross-product, which you should mention. That potentially could be more expensive.

steven-aerts: @brian-brazil  I tried to explain it more hope it is more clear.

README.md
```diff
@@ -125,6 +125,10 @@ statistics. In addition, when `aws_dimensions` is provided, the exporter needs
 to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
+In the case that all `aws_dimensions` are provided in the `aws_dimension_select` list, the exporter will not perform the
+above api request.  It will request all possible combination of values for those dimensions.
+This might reduce cost as the values for the dimensions do not need to be queried anymore, assuming that all possible value combinations are available in cloudwatch.
```
brian-brazil: might -> will

CloudWatch

available -> present

steven-aerts: Done

README.md
```diff
@@ -125,6 +125,10 @@ statistics. In addition, when `aws_dimensions` is provided, the exporter needs
 to do API requests to determine what metrics to request. This should be
 negligible compared to the requests for the metrics themselves.
 
+In the case that all `aws_dimensions` are provided in the `aws_dimension_select` list, the exporter will not perform the
+above api request.  It will request all possible combination of values for those dimensions.
+This will reduce cost as the values for the dimensions do not need to be queried anymore, assuming that all possible value combinations are present in cloudwatch.
```
brian-brazil: s/cloudwatch/CloudWatch/
s/api/API/

steven-aerts: Done

brian-brazil:
Thanks!

PR #165, opened 2019-01-22
Update HealthServlet to issue 5xx reponses when there are scrape errors

robbie-demuth:
https://github.com/prometheus/cloudwatch_exporter/pull/116 added `/-/ready` and `/-/healthy` endpoints for use in readiness and liveness probes in Kubernetes. When the exporter continuously fails to pull data from CloudWatch, the endpoints should return a non-200 response code in order to signify to Kubernetes that something may be wrong with the container. Kubernetes can then take action if the probes fail repeatedly and restart the container if necessary

@brian-brazil

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/WebServer.java
+++ b/src/main/java/io/prometheus/cloudwatch/WebServer.java
@@ -33,11 +33,10 @@ public class WebServer {
         server.setHandler(context);
         context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
         context.addServlet(new ServletHolder(new DynamicReloadServlet(collector)), "/-/reload");
-        context.addServlet(new ServletHolder(new HealthServlet()), "/-/healthy");
-        context.addServlet(new ServletHolder(new HealthServlet()), "/-/ready");
+        context.addServlet(new ServletHolder(new HealthServlet(collector)), "/-/healthy");
+        context.addServlet(new ServletHolder(new HealthServlet(collector)), "/-/ready");
         context.addServlet(new ServletHolder(new HomePageServlet()), "/");
         server.start();
         server.join();
     }
 }
-
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -32,7 +32,10 @@ import java.util.regex.Pattern;
 import org.yaml.snakeyaml.Yaml;
 
 public class CloudWatchCollector extends Collector {
+    public static final String SCRAPE_ERROR_MSG = "CloudWatch scrape failed";
+
     private static final Logger LOGGER = Logger.getLogger(CloudWatchCollector.class.getName());
+    private boolean scrapeError;
 
     static class ActiveConfig implements Cloneable {
         ArrayList<MetricRule> rules;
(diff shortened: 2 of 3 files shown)


src/main/java/io/prometheus/cloudwatch/WebServer.java
```diff
@@ -33,11 +33,10 @@ public static void main(String[] args) throws Exception {
         server.setHandler(context);
         context.addServlet(new ServletHolder(new MetricsServlet()), "/metrics");
         context.addServlet(new ServletHolder(new DynamicReloadServlet(collector)), "/-/reload");
-        context.addServlet(new ServletHolder(new HealthServlet()), "/-/healthy");
-        context.addServlet(new ServletHolder(new HealthServlet()), "/-/ready");
+        context.addServlet(new ServletHolder(new HealthServlet(collector)), "/-/healthy");
+        context.addServlet(new ServletHolder(new HealthServlet(collector)), "/-/ready");
```
robbie-demuth: For the use case I'm trying to address, I certainly need the `/-/healthy` endpoint to return a non-200 response code if there is a scrape error. I'm not sure if `/-/ready` should have the same behavior since in Kubernetes, readiness simply indicates whether the endpoints for a pod should be revealed to the service that targets the pod. Since the `/metrics` endpoint will still be up even if there is a scrape error, do you think I should the configuration of the `/-/ready` endpoint as it was?

brian-brazil:
I don't think a single scrape failure indicates an issue with the exporter. This sort of information is generally used for load balancers, which you wouldn't be using with this exporter.

Have you had cases where a particular cloudwatch exporter was broken, versus cloudwatch having an issue?

robbie-demuth:
We're currently making some modifications to the Helm chart that uses this exporter to allow use of AWS session tokens so we can rotate AWS credentials. When the credentials rotate, the Kubernetes secret is updated, but the container processes still have environment variables from the old secret. This causes scraping to fail. We'd like for this issue to propagate to one or more of the endpoints to Kubernetes knows to restart the container - at which point it will pick up the new credentials and operator normally. Do you think it'd be better to see if there is a general class of exceptions related to AWS auth issues and only issue a non-200 in that case?

brian-brazil:
If there's auth mechanisms that ec2 SD supports that this doesn't, I'll take a PR to add them.

robbie-demuth:
I think we're on different pages here. The Helm chart using this exporter injects environment variables from Kubernetes secrets. We're using AWS STS role assumption, which yields short-lived credentials that expire after a given time frame. This means that we constantly update the Kubernetes secret, but the updated environment variables don't propagate to the processes in the existing container. This means that the exporter will eventually fail to export any data from CloudWatch because the credentials will have expired. We want to catch this case and somehow get fresh credentials to the exporter. Here are a few ways we could do that:

* Expose the fact that there are issues scraping via the health endpoint. Kubernetes can then use this information to terminate and restart the container it detects that there are a certain number of consecutive scrape errors
* Update the Java code to reinitialize the AWS client if the credentials are bad. This would also require modifying the Helm chart to use an AWS config file instead of environment variables
* Update the Java code to exit when it encounters invalid AWS credentials

I elected to pursue the first option because it seems like the endpoints were added solely for consumption by Kubernetes in https://github.com/prometheus/cloudwatch_exporter/pull/116

brian-brazil:
Relying on a 3rd party to restart an entire application to keep it working in normal operations sounds like a bit of a fragile design. I'd suggest looking into another way to propagate the secrets.

robbie-demuth:
In your opinion, what should the application do when there are invalid credentials? Continue to scrape even though scraping will always fail? Would you accept a PR for the second option above?

brian-brazil:
> Continue to scrape even though scraping will always fail?

Yes.

> Would you accept a PR for the second option above?

I'd be okay with reinitialising it on every scrape, presuming there's no downside to that.

robbie-demuth:
After further investigation, it looks like AWS STS role assumption isn't going to play well with the changes we'd need to make to the Helm chart to support re-initializing the client on every scrape, so I'm going to close this PR. We'll likely build a custom Docker image with the desired functionality instead

PR #173, opened 2019-03-07
Update AWS SDK to 1.11.511

fredmtyro:
We would like to update the AWS SDK version to the current release 1.11.511 as 1.11.86 brings in httpclient 4.5.2 which has this directory traversal vulnerability: [HTTPCLIENT-1803](https://issues.apache.org/jira/browse/HTTPCLIENT-1803)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,12 +46,12 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-cloudwatch</artifactId>
-      <version>1.11.86</version>
+      <version>1.11.511</version>
     </dependency>
     <dependency>
       <groupId>com.amazonaws</groupId>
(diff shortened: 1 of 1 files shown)


fredmtyro:
@brian-brazil Sorry, I don't think I tagged you properly in this issue, and didn't have my sign off set in the initial push. Please let me know if there is anything else you need to consider this issue. I've tested the change with JDK 11 in my environment here.

brian-brazil:
I presume it still works with Java 8?

fredmtyro:
I can successfully build with Java 8 and the [release notes for the AWS SDK for Java](https://github.com/aws/aws-sdk-java#release-notes) say that it requires Java 6+, so I think there is no impact there.

brian-brazil:
Thanks!

PR #184, opened 2019-05-10
fix example.yml to use proper config names

philandstuff:
This was inconsistent with the README and the code.


philandstuff:
tagging @brian-brazil in accordance with CONTRIBUTING.md

brian-brazil:
Thanks!

PR #186, opened 2019-05-24
Use GetMetricData Rather Than GetMetricStatistics

johannes-gehrs:
- Does not save on cost. This is not possible by using GetMetricData rather than GetMetricStatistics, unless you requests many data points from the past (this would not make sense for a Prometheus Exporter).
- It is, however, a LOT faster, especially if you have many dimensions (around 4x)
- Almost all the unit tests were implementation specific and I did not come around to rewriting those. Maybe it should also be discussed what exactly the tests should test, the intent of some of the tests was somewhat unclear to me.
- I definitely think this needs some more real world experience and/or love before merging. 
- A disadvantage of this new approach is that you can no longer get the unit from CloudWatch API. The unit will thus no longer be shown in HELP text. Currently it is set to UNKNOWN, but removal should be considered.


johannes-gehrs:
Oh and I used Java 7 and 8 features. Language level was set to 6 so I changed that.

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM openjdk:11 as builder
 
 WORKDIR /cloudwatch_exporter
 ADD . /cloudwatch_exporter
-RUN apt-get -qy update && apt-get -qy install maven && mvn package && \
+RUN apt-get -qy update && apt-get -qy install maven && mvn clean package && \
```
brian-brazil: Why do you need the clean?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -4,34 +4,24 @@
...
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClient;
-import com.amazonaws.services.cloudwatch.model.Datapoint;
-import com.amazonaws.services.cloudwatch.model.Dimension;
-import com.amazonaws.services.cloudwatch.model.DimensionFilter;
-import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsRequest;
-import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsResult;
-import com.amazonaws.services.cloudwatch.model.ListMetricsRequest;
-import com.amazonaws.services.cloudwatch.model.ListMetricsResult;
-import com.amazonaws.services.cloudwatch.model.Metric;
+import com.amazonaws.services.cloudwatch.model.*;
```
brian-brazil: Please leave imports explicit

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -4,34 +4,24 @@
...
+import java.io.Reader;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 
-import org.yaml.snakeyaml.Yaml;
+import com.google.common.collect.Lists;
 
+/**
```
brian-brazil: Missing comment?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -4,34 +4,24 @@
...
-import java.util.Map;
-import java.util.HashMap;
+import java.io.Reader;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 
-import org.yaml.snakeyaml.Yaml;
+import com.google.common.collect.Lists;
```
brian-brazil: Do we really need this?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -377,120 +378,155 @@ private String help(MetricRule rule, String unit, String statistic) {
...
+      return new MetricDataQuery()
+              .withId(queryId(rule, dimensions, stat))
+              .withMetricStat(metricStat);
+    }
+
+    private static List<MetricDataQuery> queriesForRuleDimensions(MetricRule rule, List<Dimension> dimensions) {
+      List<MetricDataQuery> queries = new ArrayList<MetricDataQuery>();
+
+      for (String stat : rule.allStats()) {
+        queries.add(metricDataQuery(rule, dimensions, stat));
```
brian-brazil: Presently we get all the stats as one request, could this blow up costs by up to 5x if all stats are requested?

Dockerfile
johannes-gehrs: I don't really know why but the Docker build seemed to sometimes use outdated code when I did not use "clean". We can remove this though, I kind of accidentally committed this.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
johannes-gehrs: done.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
johannes-gehrs: No, I verified with AWS support and tested. Cost stays the same.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
johannes-gehrs: We use it for exactly one call - partition a list into chunks of 100 elements. This is the maximum queries per GetMetricData request. I think it is convenient not to implement this ad hoc with a for loop and if statement.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
johannes-gehrs: Removed

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
brian-brazil: Can you add a comment to that effect, and update the README?

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
brian-brazil: I'd rather not pull in a new dependency for the sake of a few lines of code.

iyesin:
@brian-brazil what prevents this PR from being merged?

brian-brazil:
There's outstanding comments.

iyesin:
@brian-brazil I see only one open question (not change request) from you.

johannes-gehrs:
Since my org are mostly moving further away from exporting CloudWatch metrics due to the cost associated with it, and since this code would need quite a bit more work in my estimation I have decided not to pursue this further at the moment. Do feel free to iterate on the status quo.

mghildiy:
I have fetched this branch. Going by discussion here, two more changes are needed:
1. Get rid of org.yaml.snakeyaml.Yaml
2. Add comment and update  README as per comment above

Please confirm, and I would make changes and commit.

brian-brazil:
That sounds about right.

mghildiy:
Added to 'Cost' section in README as:

```
We are using AWS Cloudwatch SDK's GetMetricDataRequest to collect the statistics. It is tested that cost remains same
even if request stats in multiple calls.
```

brian-brazil:
It'd be best to open a PR for that detail of review.

ofiry-tm:
hey, what heppened with this PR? 
the scraping is very slow, is there another solution for the slowness ?

matthiasr:
We didn't quite finish it due to bandwidth. If you want to pick it up, use this branch as a starting point, resolve any merge conflicts, and go through the comments here for other outstanding items. In principle, this is still what we want!

PR #191, opened 2019-06-14
Switch to Alpine and bump to Java 13

S-Corkum:
The current base image which is debian based has several CVE's listed as High. Changing to the Alpine image (which also requires changing to Java 13) remediates the security vulnerabilities, allowing the image to pass Clair scanning


brian-brazil:
Let's stick with a LTS release.

PR #192, opened 2019-06-20
Close the FileReader

rnzsgh:
The purpose of this PR is to close the FileReader after use. If the reloadConfig method is called excessively, it could potentially leak a number of FDs.


brian-brazil:
Can you add the DCO please?

PR #195, opened 2019-07-23
Improve build process

iyesin:
There are two improvements in this PR:
- Decrease build time: since we are using `builder` intermediate container we can skip cleanup step
- Decrease amount of meaningless messages during build process

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM openjdk:11 as builder
 
 WORKDIR /cloudwatch_exporter
 ADD . /cloudwatch_exporter
-RUN apt-get -qy update && apt-get -qy install maven && mvn package && \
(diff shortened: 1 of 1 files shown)


brian-brazil:
From a debugging standpoint, I think we should keep the messages

iyesin:
> From a debugging standpoint, I think we should keep the messages

Errors, exceptions and warnings are still shown. From my perspective I don't see any help from verbose messages during image build progress (except making it hard to notice warnings among progress-like messages).

brian-brazil:
I'd prefer to see the same output as when I do a mvn package by hand.

iyesin:
@brian-brazil what form of output? With cache present or without it?

brian-brazil:
The output from mvn

iyesin:
@brian-brazil hope this will please you.

brian-brazil:
Thanks!

PR #204, opened 2019-08-29
Adding extra verbosity/detail into README

pmiles:
@brian.brazil@robustperception.io

A small change to the README file to make it easier for Docker noobs to get going.


brian-brazil:
This more is basic docker documentation than exporter documentation, it'd probably be best to put up on a blog or similar.

PR #205, opened 2019-09-02
Fix metric loading for AWS/ES

alexnederlof:
AWS ElasticSearch does not return any metrics when the client ID and the cluster are not set. This PR takes the `awsDimensionSelect` and puts that into the dimension filters with the values, so the metrics will actually return a value. 

This is slightly brittle, but so is the AWS API apparently

Let me know what you think @brian-brazil

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -259,8 +259,16 @@ public class CloudWatchCollector extends Collector {
       request.setNamespace(rule.awsNamespace);
       request.setMetricName(rule.awsMetricName);
       List<DimensionFilter> dimensionFilters = new ArrayList<DimensionFilter>();
-      for (String dimension: rule.awsDimensions) {
-        dimensionFilters.add(new DimensionFilter().withName(dimension));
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'd prefer that this is dealt with by users in the config file, I'd prefer to only add workarounds for things that can't be handled any other way.

alexnederlof:
I agree, but that's not an option here. Without the value in the `dimensionFilters`, no dimensions are returned. Without dimensions, the metric does not return any values. `AWS/ES` is different/inconsistent in that regard from the other APIs from Amazon.

brian-brazil:
There's other services that require dimensions, this isn't unusual.

alexnederlof:
So how would you like to proceed? As far as I can tell, `AWS/ES` will not work without changes to the code base. Are you willing to support `AWS/ES` and make some changes, or do you want to prevent exceptions from entering, and drop support for `AWS/ES`?

brian-brazil:
To me it looks like your `aws_dimensions` haven't been specified properly.

alexnederlof:
You're absolutely right! I had a case sensitive typo (`ClientId` vs `ClientID`) that prevented the dimensions from overlapping correctly 🤦.

Thanks for your help!

PR #216, opened 2019-10-10
Move away from deprecated SDK methods and cleanup imports

louisfelix:
Hi @brian-brazil! This PR is about these 2 items:
- Move away from deprecated sdk methods. Doing so, use the now standard builder pattern. In the case of AmazonCloudWatch, it now works with interface instead of direct implementation.
- Remove unused imports

Thanks


brian-brazil:
Thanks!

PR #219, opened 2019-10-17
Close the FileReader

asherf:
(also sorts imports)

This is basically https://github.com/prometheus/cloudwatch_exporter/pull/192 
but I have DCO signed... so this can get merged.


brian-brazil:
Thanks!

PR #221, opened 2019-10-23
add version, help, and debug flag, debug log GetMetricStatisticsResults

dt-rush:
Change-type: minor
Signed-off-by: dt-rush <nickp@balena.io>

PR #222, opened 2019-10-23
add version, help, and debug flag, debug logging (no apache commons CLI dependency)

dt-rush:
Change-type: minor
Signed-off-by: dt-rush <nickp@balena.io>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,7 @@
           <archive>
             <manifest>
               <mainClass>io.prometheus.cloudwatch.WebServer</mainClass>
+              <addDefaultImplementationEntries>true</addDefaultImplementationEntries> 
             </manifest>
           </archive>
           <descriptorRefs>
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 import org.yaml.snakeyaml.Yaml;
 
 public class CloudWatchCollector extends Collector {
-    private static final Logger LOGGER = Logger.getLogger(CloudWatchCollector.class.getName());
+    private static final Logger LOGGER = Logger.getLogger(CloudWatchCollector.class.getPackage().getName());
 
     static class ActiveConfig implements Cloneable {
         ArrayList<MetricRule> rules;
@@ -60,6 +60,10 @@ public class CloudWatchCollector extends Collector {
       Map<String,List<String>> awsDimensionSelectRegex;
       String help;
       boolean cloudwatchTimestamp;
+
+      public String toString() {
+        return String.format("MetricRule<namespace=%s, metricName=%s, periodSeconds=%d, rangeSeconds=%d, delaySeconds=%d, statistics=%s, extendedStatistics=%s, dimensions=%s, cloudwatchTimestamp=%b", this.awsNamespace, this.awsMetricName, this.periodSeconds, this.rangeSeconds, this.delaySeconds, this.awsStatistics, this.awsExtendedStatistics, this.awsDimensions, this.cloudwatchTimestamp);
+      }
     }
 
     ActiveConfig activeConfig = new ActiveConfig();
@@ -209,8 +213,8 @@ public class CloudWatchCollector extends Collector {
           } else {
               rule.cloudwatchTimestamp = defaultCloudwatchTimestamp;
           }
+          LOGGER.log(Level.FINEST, String.format("Parsed rule: %s", rule));
         }
-
         loadConfig(rules, client);
     }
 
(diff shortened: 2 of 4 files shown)


brian-brazil:
I'm okay with the additional logging, but the rest isn't needed.

dt-rush:
@brian-brazil a help message which shows the flags properly, and a version flag, are just basic features of a CLI which are currently missing. Would you like me to separate those changes into a separate PR after this one adding the debug flag? Or we could just leave them in this one.

brian-brazil:
There's no need for flags for something as simple as this, and the version would be better on the /metrics.

dt-rush:
@brian-brazil Ah, adding the version in `/metrics` is a good point. This is common among exporters. But is there any performance reason or design principle involved in not supporting universal CLI paradigms of `-h / --help` and `-v / --version`?

brian-brazil:
It's extra complexity not needed for something that takes two arguments.

dt-rush:
The thing is, every other exporter hosted by the `prometheus` github organization supports `-h` and/or `--help`. I don't see this as a meaningful use of the word "complexity" in the space of software design.

- [`node_exporter`](https://github.com/prometheus/node_exporter#building-and-running)
- [`consul_exporter`](https://github.com/prometheus/consul_exporter#flags)
- [`statsd_exporter`](https://github.com/prometheus/statsd_exporter#building-and-running) 
- [`haproxy_exporter`](https://github.com/prometheus/haproxy_exporter#getting-started)
- [`influxdb_exporter`](https://github.com/prometheus/influxdb_exporter/blob/master/main.go#L266) (the README doesn't state this)
- [`graphite_exporter`](https://github.com/prometheus/graphite_exporter/blob/master/main.go#L249) (the README doesn't state this) 
- [`collectd_exporter`](https://github.com/prometheus/collectd_exporter/blob/master/main.go#L315) (the README doesn't state this) 
- [`mysqld_exporter`](https://github.com/prometheus/mysqld_exporter/blob/master/mysqld_exporter.go#L247) (the README doesn't state this)
- [`memcached_exporter`](https://github.com/prometheus/memcached_exporter) (the README doesn't state this)
- [`blackbox_exporter`](https://github.com/prometheus/blackbox_exporter#running-this-software) (the README doesn't state this)
- [`snmp_exporter`](https://github.com/prometheus/snmp_exporter) (the README doesn't state this)

brian-brazil:
Java exporters don't use the Kingpin library that the Go ones do.

dt-rush:
**Comparison of Go and Java argument parsing** 

LOC is directly comparable between go and Java in this case. Compare:

```{go}
// help
var (
		listenAddress             = kingpin.Flag("web.listen-address", "Address to listen on for web interface and telemetry.").Default(":9101").String()
		metricsPath               = kingpin.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").String()
		haProxyScrapeURI          = kingpin.Flag("haproxy.scrape-uri", "URI on which to scrape HAProxy.").Default("http://localhost/;csv").String()
		haProxySSLVerify          = kingpin.Flag("haproxy.ssl-verify", "Flag that enables SSL certificate verification for the scrape URI").Default("true").Bool()
		haProxyServerMetricFields = kingpin.Flag("haproxy.server-metric-fields", "Comma-separated list of exported server metrics. See http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#9.1").Default(serverMetrics.String()).String()
		haProxyTimeout            = kingpin.Flag("haproxy.timeout", "Timeout for trying to get stats from HAProxy.").Default("5s").Duration()
		haProxyPidFile            = kingpin.Flag("haproxy.pid-file", pidFileHelpText).Default("").String()
	)

...

// parsing
	flag.AddFlags(kingpin.CommandLine, promlogConfig)
	kingpin.HelpFlag.Short('h')
	kingpin.Parse()
```

and the java:

```{java}
// help
private static void printHelp() {
        System.out.println("cloudwatch_exporter [OPTIONS] <port> <yml configuration file>");
        System.out.println("-d,--debug     enable debug logs");
        System.out.println("-h,--help      print this message");
        System.out.println("-v,--version   print version");
    }

...

// parsing
private static String[] parseOptions(String[] args) {
      int highestOptionIndex = 0;
      for (String arg : args) {
        if (arg.equals("-h") || arg.equals("--help")) {
          printHelp();
          System.exit(0);
        }
        else if (arg.equals("-v") || arg.equals("--version")) {
          printVersion();
          System.exit(0);
        }
        else if (arg.equals("-d") || arg.equals("--debug")) {
          enableDebugLogs();
          highestOptionIndex++;
        }
      }
      return Arrays.copyOfRange(args, highestOptionIndex, args.length);
    }
```

The [original version of this PR](https://github.com/prometheus/cloudwatch_exporter/pull/221) in fact used a [small library](https://mvnrepository.com/artifact/commons-cli/commons-cli/1.4) compare with kingpin (and only easier to read and maintain, but not more or less "complex" in LOC (not a true measure of complexity) than the current pure Java version), but per your comments, we switched to pure Java, although the difference in the built jar is in fact only a few bytes.

**Keeping it simple**

I understand and fully agree that we should keep exporters simple. I don't see this as meaningfully increasing the complexity of the exporter, only keeping parity with basic CLI features available in virtually every binary produced in the modern era.

**Next steps**

If you stand firm on your current stand, of course as a community contributor I have no recourse but to remove the flags, so as to enable the PR to pass, and achieve meaningful debug logging in this component. I will say that the least we can ask of one another in this public collaborative venture of prometheus is that we use our reasoning to find our way to improvements, and I find the reasoning in this case - for not including `-h` and/or `-v` - to be thin. 

So, let me know if we can pass this PR as is, or whether I'll take out the `-h` / `-v` (making this the odd one out among prometheus exporters), and leave those as an issue on this repo for posterity to consider.

PR #224, opened 2019-10-24
Update AWS SDK to 1.11.658

willthames:
Using IAM Service roles for EKS needs a newer version of the SDK.
Updating to the latest at the time of writing

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,12 +46,12 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-cloudwatch</artifactId>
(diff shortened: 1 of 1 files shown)


willthames:
@brian-brazil

brian-brazil:
Thanks!

lachlancooper:
IAM Roles for Service Accounts are not currently working (at least without blocking the metadata API) because in the Java SDK the priority of the web identity token provider is below the instance profile provider.

This should be fixed soon: https://github.com/aws/aws-sdk-java/issues/2136

Once released we'll need to update the SDK version again.

PR #227, opened 2019-10-25
Only close file if we opened a file.

asherf:
I noticed when testing stuff locally, that if I point it to a file that doesn't exist calling `reader.close();` causes an error (NullPointerException) which hides the original error (FileNotFoundException)
This fixes that situation.


brian-brazil:
Thanks!

PR #228, opened 2019-10-29
Support the configuration of the exporter based on tag name and tag values

### LINKED ISSUE #96: Add optional support for ec2 / elb / rds / es filtering of tags and tag labels
Hey Ho,

i would like to implement filtering / target discovery based on tags for a small list of aws services. The tags would be added as prometheus labels as well.

Saw a closed issue (https://github.com/prometheus/cloudwatch_exporter/issues/20) where this statement where added - "By my count there are currently 41 AWS services (and that's just AWS services) providing data in Cloudwatch. We're not going to add every possible object field from those." - Personally i really love this project but we quite use it only for old static things which is sad. New things e.g. elbs managed by kubernetes are not monitorable with this project as ids change a lot (tags are the same). Even if we would write something which automatically generates a config for this project, the exported metrics would not help because of missing tag labels. Therefore i would like to start working on this problem.

There would be a 2nd option to implement this - Giving users the ability to add custom labels to exported metrics. Then we could write a sidecar app which just creates/manages the config file and reloads exporter on changes.

From my perspective i prefer option one as it allows users easy monitoring of aws resources without running a 2nd container.

I would start on of those two options as soon as i get your go!

Greetings, Thomas

louisfelix:
Hi @brian-brazil ,

This PR fixes #96 . As discussed with you in the latest issue comments, I am adding the possibility to configure the exporter based on resource tag name / tag values. 

The new optional `aws_tag_select` property is compatible with the existing `aws_dimensions` property, and is applied similarly to `aws_dimension_select` or `aws_dimension_select_regex`.

Thanks!


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -304,10 +392,10 @@ public String getMonitoringEndpoint(Region region) {
     }
 
     /**
-     * Check if a metric should be used according to `aws_dimension_select` or `aws_dimension_select_regex`
+     * Check if a metric should be used according to `aws_dimension_select`, `aws_dimension_select_regex` or dynamic `aws_tag_select`
      */
-    private boolean useMetric(MetricRule rule, Metric metric) {
-      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null) {
+    private boolean useMetric(MetricRule rule, List<String> tagBasedResourceIds, Metric metric) {
```
brian-brazil: I'd expect this to be an and, not an or

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -130,28 +147,26 @@ private void loadConfig(Map<String, Object> config, AmazonCloudWatch client) {
...
           }
+          Region region = getRegion(config);
           clientBuilder.setEndpointConfiguration(new EndpointConfiguration(getMonitoringEndpoint(region), region.getName()));
 
-          client = clientBuilder.build();
+          cloudWatchClient = clientBuilder.build();
+        }
+        
+        if (taggingClient == null) {
+          AWSResourceGroupsTaggingAPIClientBuilder clientBuilder = AWSResourceGroupsTaggingAPIClientBuilder.standard();
```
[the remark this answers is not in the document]
louisfelix: No it won't. All the `loadConfig` is even successful without internet connection.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
louisfelix: Currently, the logic is an `or` between aws_dimension_select and aws_dimension_select_regex.
Changing this to an `and` while adding aws_tag_select would break existing client behavior if they already use both aws_dimension_select and aws_dimension_select_regex in a same config bloc.
Do you think it worth changing to an `and` ?

brian-brazil: It doesn't make sense to use both aws_dimension_select and aws_dimension_select_regex, so I'd go for and.

README.md
```diff
@@ -22,9 +22,12 @@ This includes the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment
 variables.
 
 The `cloudwatch:ListMetrics` and `cloudwatch:GetMetricStatistics` IAM permissions are required.
+The `tag:GetResources` IAM permission is also required to use the `aws_tag_select` configuration.
```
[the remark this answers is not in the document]
louisfelix: Fixed!

README.md
```diff
@@ -46,6 +67,10 @@ aws_metric_name  | Required. Metric name of the CloudWatch metric.
 aws_dimensions | Optional. Which dimension to fan out over.
 aws_dimension_select | Optional. Which dimension values to filter. Specify a map from the dimension name to a list of values to select from that dimension.
 aws_dimension_select_regex | Optional. Which dimension values to filter on with a regular expression. Specify a map from the dimension name to a list of regexes that will be applied to select from that dimension.
+aws_tag_select | Optional. A tag configuration to filter on, based on mapping from the tagged resource ID to a CloudWatch dimension.  
+tag_selections | Required under `aws_tag_select`. Specify a map from a tag key to a list of tag values to apply [tag filtering](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters) on resources from which metrics will be gathered.
+resource_type_selection | Required under `aws_tag_select`. Specify the [resource type](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-ResourceTypeFilters) to filter on.
```
brian-brazil: https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-aws-service-namesspaces would be more useful I suspect.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -304,19 +392,23 @@ public String getMonitoringEndpoint(Region region) {
...
-      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null) {
+    private boolean useMetric(MetricRule rule, List<String> tagBasedResourceIds, Metric metric) {
+      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null && rule.awsTagSelect == null) {
         return true;
-      }
-      if (rule.awsDimensionSelect != null  && metricsIsInAwsDimensionSelect(rule, metric)) {
-        return true;
-      }
-      if (rule.awsDimensionSelectRegex != null  && metricIsInAwsDimensionSelectRegex(rule, metric)) {
+      } else {
```
brian-brazil: You can simplify here. The body of this else is enough.

louisfelix: @brian-brazil Oh... indeed! I fixed it.

src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -304,19 +392,22 @@ public String getMonitoringEndpoint(Region region) {
     }
 
     /**
-     * Check if a metric should be used according to `aws_dimension_select` or `aws_dimension_select_regex`
+     * Check if a metric should be used according to `aws_dimension_select`, `aws_dimension_select_regex` and dynamic `aws_tag_select`
      */
-    private boolean useMetric(MetricRule rule, Metric metric) {
-      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null) {
+    private boolean useMetric(MetricRule rule, List<String> tagBasedResourceIds, Metric metric) {
+      if (rule.awsDimensionSelect == null && rule.awsDimensionSelectRegex == null && rule.awsTagSelect == null) {
```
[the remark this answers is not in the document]
louisfelix: @brian-brazil  Indeed again. I fixed it.

brian-brazil:
Thanks!

PR #229, opened 2019-11-06
Support resources tags as metric label

louisfelix:
Hi @brian-brazil,

As discussed with you in the latest #96 issue comments (item 2. : "if we happen to have the metadata for it"), this PR adds the resources tags as metric label in a new metric called `aws_resource_info`. This metric contains tags as labels for resources selected through `aws_tag_select`, and can later be joined to other metrics through PromQL queries.

Note that I have kept the empty `instance` label, even though I am not sure of the reason it is there, both in the case of the new metric and in other existing metrics. I'm curious to have your comment on that.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,10 +40,11 @@ metrics:
    aws_statistics: [Sum]
 ```
 
-A similar example with common options and `aws_tag_select`:
+A similar example with common options, `aws_tag_select` and `enable_aws_resource_info`:
 ```
 ---
 region: eu-west-1
+enable_aws_resource_info : true
 metrics:
  - aws_namespace: AWS/ELB
    aws_metric_name: RequestCount
@@ -68,9 +69,10 @@ aws_dimensions | Optional. Which dimension to fan out over.
 aws_dimension_select | Optional. Which dimension values to filter. Specify a map from the dimension name to a list of values to select from that dimension.
 aws_dimension_select_regex | Optional. Which dimension values to filter on with a regular expression. Specify a map from the dimension name to a list of regexes that will be applied to select from that dimension.
 aws_tag_select | Optional. A tag configuration to filter on, based on mapping from the tagged resource ID to a CloudWatch dimension.  
-tag_selections | Required under `aws_tag_select`. Specify a map from a tag key to a list of tag values to apply [tag filtering](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters) on resources from which metrics will be gathered.
-resource_type_selection | Required under `aws_tag_select`. Specify the [resource type](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-aws-service-namesspaces) to filter on.
-resource_id_dimension | Required under `aws_tag_select`. For the current metric, specify which CloudWatch dimension maps to the ARN [resource ID](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax).
+tag_selections | Required, under `aws_tag_select`. Specify a map from a tag key to a list of tag values to apply [tag filtering](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters) on resources from which metrics will be gathered.
+resource_type_selection | Required, under `aws_tag_select`. Specify the [resource type](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-aws-service-namesspaces) to filter on.
+resource_id_dimension | Required, under `aws_tag_select`. For the current metric, specify which CloudWatch dimension maps to the ARN [resource ID](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax).
+enable_aws_resource_info | Optional, under `aws_tag_select`. Boolean for whether to enable the export of `aws_resource_info` metric from tagging API results. Defaults to false. Can be set globally and per `aws_tag_select`.
 aws_statistics | Optional. A list of statistics to retrieve, values can include Sum, SampleCount, Minimum, Maximum, Average. Defaults to all statistics unless extended statistics are requested.
 aws_extended_statistics | Optional. A list of extended statistics to retrieve. Extended statistics currently include percentiles in the form `pN` or `pN.N`.
 delay_seconds | Optional. The newest data to request. Used to avoid collecting data that has not fully converged. Defaults to 600s. Can be set globally and per metric.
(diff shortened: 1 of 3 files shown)


brian-brazil:
Users should be able to opt in for this explicitly, rather than depending on using a specific feature that they may not need.

Yes empty instance should stay, it's so the metrics can end up without the exporters instance label using honor_labels.

louisfelix:
@brian-brazil Ok got it for the `instance`.

To address your concern, here is a new config proposition that would support explicit user opt-in, remove dependency on the other feature, but still leverage [GetResource API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html) and `aws_tag_select` logic. I would basically separate this as an optional config section called `aws_resource_info_metrics` (or maybe just `info_metrics` ?) to manage the info metrics:

```
metrics:
...

aws_resource_info_metrics:
- aws_tag_select:
    resource_type_selection: "elasticloadbalancing:loadbalancer/app"
    resource_id_dimension: LoadBalancer
    tag_selections:
      Monitoring: ["enabled"]
- aws_tag_select:
    resource_type_selection: "ec2:instance"
    resource_id_dimension: InstanceId
    tag_selections:
      Env: ["prod", "uat"]
```

Let me know what you think!

brian-brazil:
I'd have the option as a simple bool, it's a permissions questions more than anything else.

louisfelix:
@brian-brazil Ok good. I went ahead and changed the PR by adding the option to opt-in.

README.md
```diff
@@ -68,9 +69,10 @@ aws_dimensions | Optional. Which dimension to fan out over.
...
+tag_selections | Required, under `aws_tag_select`. Specify a map from a tag key to a list of tag values to apply [tag filtering](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters) on resources from which metrics will be gathered.
+resource_type_selection | Required, under `aws_tag_select`. Specify the [resource type](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-aws-service-namesspaces) to filter on.
+resource_id_dimension | Required, under `aws_tag_select`. For the current metric, specify which CloudWatch dimension maps to the ARN [resource ID](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax).
+enable_aws_resource_info | Optional, under `aws_tag_select`. Boolean for whether to enable the export of `aws_resource_info` metric from tagging API results. Defaults to false. Can be set globally and per `aws_tag_select`.
```
brian-brazil: This should only be a global.

README.md
```diff
@@ -82,8 +84,15 @@ The above config will export time series such as
 ```
 # HELP aws_elb_request_count_sum CloudWatch metric AWS/ELB RequestCount Dimensions: ["AvailabilityZone","LoadBalancerName"] Statistic: Sum Unit: Count
 # TYPE aws_elb_request_count_sum gauge
-aws_elb_request_count_sum{job="aws_elb",load_balancer_name="mylb",availability_zone="eu-west-1c",} 42.0
-aws_elb_request_count_sum{job="aws_elb",load_balancer_name="myotherlb",availability_zone="eu-west-1c",} 7.0
+aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="mylb",availability_zone="eu-west-1c",} 42.0
+aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="myotherlb",availability_zone="eu-west-1c",} 7.0
+```
+
+If `enable_aws_resource_info` is set to true, an additional information metric will be exported for each AWS resource matched by `aws_tag_select`, such as
```
brian-brazil: It should apply to everything.

louisfelix: @brian-brazil I know, but as you told the opt-in was "a question of permission", I though you mean the boolean would be enough.

I would have like to apply it to everything, but the dependency to the tag selection is here because the necessary meta data is received through the [GetResource API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html). The API takes type and tag filters, and returns all tags mapped to an ARN; this result can't be generically correlated to the CloudWatch dimensions, as the available dimensions may or may not contain the resource_id that we would extract from the ARN (and we would have to know which dimension represents the resource_id).

So, I saw 3 options
1) The info metric depend on the tag selection feature usage (this initial PR)
2) The info metric is configured separately in a dedicated config bloc, so it's a way to both opt-in and remove dependency to the tag selection. Finally, this may be a way more intuitive solution (proposed in my [comment here](https://github.com/prometheus/cloudwatch_exporter/pull/229#issuecomment-551208385))
3) The info metric is transparently available to any metric selection, through an opt-in boolean. The problem here is, unless I miss something, we would have to add a dependency to each resource type SDK, and call and maintain specific code for each (e.g. building an `AmazonEC2` client, building a `DescribeTagsRequest()`, etc..) and even than, any correlation code with CloudWatch dimension would need to be hardcoded. I don't see it as an acceptable or maintainable solution.

Let me know what you think!

brian-brazil: > but the dependency to the tag selection is here because the necessary meta data is received through the

Ah yes.

> The info metric depend on the tag selection feature usage (this initial PR)

This would be the way to go I think, which means tag_selections should be optional.

louisfelix: > This would be the way to go I think, which means tag_selections should be optional.

@brian-brazil . Good! So in summary:
- The tag selection mechanism (`aws_tag_select`) is already optional, so nothing to change here
- Do you still think we need a global only `enable_aws_resource_info` boolean, or should I just remove this boolean?

louisfelix:
> We should also add to the costing docs/metrics.
This PR does not add any API call, so I think we are OK.

README.md
brian-brazil: > is already optional, so nothing to change here

The docs indicate otherwise.

> or should I just remove this boolean?

We can remove it.

brian-brazil:
> This PR does not add any API call, so I think we are OK.

We should still add them.

README.md
louisfelix: > The docs indicate otherwise.

The doc say "aws_tag_select | Optional. "
But you may refer to the tag_selection sub-element, which as changed in this PR because I added a coma.
The doc says: tag_selections | Required under `aws_tag_select`, maybe it's not clear enough, should I say instead:
"tag_selections | Required when using `aws_tag_select`"

louisfelix:
> We should still add them.

Ok. But the tag feature does not add CloudWatch API call, so I should not update `cloudwatch_requests_total`. It is calling the [Resource Groups Tagging API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/Welcome.html) , but even though I could not find cost associated to that, should I add a new metric counting these?

README.md
brian-brazil: tag_selections should no longer be required, as it's not needed if you just want this new metric.

brian-brazil:
https://aws.amazon.com/blogs/aws/new-aws-resource-tagging-api/ indicates they're free, but we should still count them and mention that they're free.

louisfelix:
> https://aws.amazon.com/blogs/aws/new-aws-resource-tagging-api/ indicates they're free, but we should still count them and mention that they're free.

Ok!

README.md
louisfelix: Ah! yes ok.

louisfelix:
Hi @brian-brazil , ok so I made the changes: removed `enable_aws_resource_info` boolean, made `tag_selections` optional, and added cost doc/metric (`tagging_api_requests_total`).

brian-brazil:
Thanks!

PR #233, opened 2019-11-19
Return tag _info when there's no tag_selections

brian-brazil:
@louisfelix

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -475,6 +475,9 @@ public class CloudWatchCollector extends Collector {
      * Check if a metric is matched in `aws_tag_select`
      */
     private boolean metricIsInAwsTagSelect(MetricRule rule, List<String> tagBasedResourceIds, Metric metric) {
+      if (rule.awsTagSelect.tagSelections == null) {
+        return true;
+      }
       for (Dimension dimension : metric.getDimensions()) {
         String dimensionName = dimension.getName();
         String dimensionValue = dimension.getValue();


louisfelix:
@brian-brazil , I think it can't hurt, but in the tests I did (and in the unit test I added for this case), this additional check was not needed as when there is no `tag_selection`, the `tagBasedResourceIds` list will contain all the resources of the given types (all the resources that have tags or _had_ tags are returned, as per [API specifications](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters), so all resources that would benefit from the _info metric would be in the list), so it returns true at the end of the dimension looping.

Am I missing something (or did you see a case where it was needed) ?

brian-brazil:
When I tested it out, this didn't work for me without it.

louisfelix:
ok, so I must be doing a wrong assumption here.
LGTM then

louisfelix:
Thanks! ;-)

louisfelix:
@brian-brazil I am still curious to know the test case, if you have it handy... Thanks!

brian-brazil:
I tried it on a random S3 bucket.

brian-brazil:
> all the resources that have tags or had tags are returned

Which is not the same as all resources.

louisfelix:
> Which is not the same as all resources.

Yes exactly. I am doing more tests now (with random S3 also), an _info metric is added for each tagged/previously tagged resources. I don't think your PR change this behaviour.

This was the expected behaviour, as the goal here is to provide tags as label, this the useful info the _info metric provides.

The README also says that: "an additional information metric will be exported for each AWS **tagged** resource"

brian-brazil:
Yes, but it breaks the existing metrics for anything that has never had a tag.

louisfelix:
> Yes, but it breaks the existing metrics for anything that has never had a tag.

Ahhhhh. Yes true, and indeed your PR is fixing that (I tested _before_ and _after_ your PR, with an EC2 that has never had tag).

I could enhance unit tests after your PR is merged, to cover this case.

PR #234, opened 2019-11-21
Enhance unit test for resource with not tag when tag_selection is not specified

louisfelix:
Hi @brian-brazil ,
As discussed in PR #233 , I'm adding here an enhancement to the unit test to cover the case of a resource with not tag when `tag_selection` is not specified under `aws_tag_select`.
The enhanced test fail without PR #233 , but succeed with it, as expected.
Thanks!


brian-brazil:
Thanks!

PR #238, opened 2019-12-06
Add resource tags to existing metrics

Chrisss93:
#229 adds AWS resource tag key-values in the `aws_resource_info` metric. I think it might be  convenient to add the resource tags to the existing metrics based on the matching resourceARN if `aws_tag_select` option is configured. This would make it easier for downstream prometheus querying as you have access not only to an AWS metric's dimension, but the tags of its chosen resource dimension for a richer label set. I rarely write Java, so let me know if this is up to snuff.

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -575,6 +575,16 @@ public class CloudWatchCollector extends Collector {
           for (Dimension d: dimensions) {
             labelNames.add(safeName(toSnakeCase(d.getName())));
             labelValues.add(d.getValue());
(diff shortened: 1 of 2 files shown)


brian-brazil:
These are basically target labels so https://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas applies, so this doesn't make sense in general. You can join as needed in PromQL.

fblanke:
This PR was created some while ago, but I too would like this functionality.

Basically we share a prometheus instance across many of our teams, and with this feature, we would be able to separate metrics specifically on SQS with a tag like `team: team-1` etc which would be helpful when alerting. Other than that, the only dimension available is Queuename which is not very helpful. So this would be a good merge @brian-brazil

brian-brazil:
For the reasons I've already given this id not in line with label best practices, so this is unlikely to be merged. An info metric as we already expose this information is the right way to do this.

PR #239, opened 2019-12-10
Support default provider chain for region

alkalinecoffee:
Adds support for falling back to the default region provider chain to fetch the region if omitted from the configuration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ metrics:
 
 Name     | Description
 ---------|------------
-region   | Optional. The AWS region to connect to. If none is provided, the region from the instance metadata is used.
+region   | Optional. The AWS region to connect to. If none is provided, an attempt will be made to determine the region from the [default region provider chain](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html#default-region-provider-chain).
 role_arn   | Optional. The AWS role to assume. Useful for retrieving cross account metrics.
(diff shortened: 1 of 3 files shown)


brian-brazil:
We only support whatever the libraries we use happen to support, and having multiple places a config option can come from is confusing.

alkalinecoffee:
Slept on this a bit and just pushed some changes.

Unless explicitly set, most, if not all AWS SDKs have support for deriving the region from other sources, such as environment variables or AWS config files, in addition to the EC2 metadata endpoint.

This PR has been updated to follow common patterns set by AWS SDKs and CLI tools, where the SDK silently falls back to the provider chain to determine the desired region.

See this for a list of how the SDK will determine the region:

https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html#default-region-provider-chain

I originally intended to provide support just for the `AWS_DEFAULT_REGION` envvar, which as far as I know is the newer and preferred region envvar, but the Java SDK continues to support only the older `AWS_REGION` envvar.  Either way, this updated solution is more flexible and more in-line with patterns established by AWS.

brian-brazil:
That looks better, I presume this will work for things like gov and China?

alkalinecoffee:
Since we're relying on the SDK here, I can't imagine there would be an issue.  But I've no access nor experience in those environments so I'm unable to validate myself.

brian-brazil:
Thanks!

PR #245, opened 2020-01-07
[Issue-241] Remove ':' from safeName regex pattern

ccjenniferhsu:
Currently, the metrics label name can contain ":" from the exporter, This doesn't match prometheus regex (See [official doc]( https://prometheus.io/docs/concepts/data_model/)), the regex should be `[a-zA-Z0-9_]` In this PR, we change the regex in the safeName function which change the invalid chars to underscore.


brian-brazil:
Can you add the DCO?

Mattias-:
Metric names may actually include `:`. This is changing those metric names and is not backwards compatible.

brian-brazil:
Just saw you'd updated this now, #247 does this in a slightly better way so I've taken that one. Thanks for your PR!

PR #247, opened 2020-01-15
Fix label name strictness

### LINKED ISSUE #241: Exporter will be broken if using aws_tag_select to specify the resources and there are ":" inside the label
When using `aws_tag_select` to select resources, it will generates `aws_resource_info` metrics with labels like `tag_$KeyName`. However, the exporter will be invalid for prometheus if  we select cloudformation based resource due to invalid label name.

Here is one example:

We used cloudformation to generate one RDS with stack name "grafanaRDS". According to [AWS resource tag](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html), it will have few tags got generated accordingly, and `aws:cloudformation:stack-name` is one of them. 

Next, I use `aws:cloudformation:stack-name` in the configuration:

>   - aws_namespace: AWS/RDS
        aws_metric_name: BinLogDiskUsage
        aws_dimensions: [DBInstanceIdentifier]
        aws_statistics: [Average]
        aws_tag_select:
          tag_selections:
            tag_aws:cloudformation:stack_name: ["grafanaRDS"]
          resource_type_selection: "rds:db"
          resource_id_dimension: DBInstanceIdentifier

From metrics page, it does generate the following metrics:
>aws_resource_info{job="aws_rds",instance="",arn="...",dbinstance_identifier="...",tag_Description="Grafana RDS",tag_aws:cloudformation:logical_id="DBGrafana",tag_aws:cloudformation:stack_id="...",tag_aws:cloudformation:stack_name="grafanaRds",} 1.0`

However, prometheus can't scrape this target correctly and is reporting `INVALID` error: 
Screen Shot 2019-12-13 at 2 15 27 PM copy

The reason is the generated labels doesn't match [label regex](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) : `[a-zA-Z_][a-zA-Z0-9_]*` 

Since people might use cloudformation to generate AWS resource frequently, it would be nice if the exporter can support `aws_tag_select` feature with these. Could we fix this bug? Thanks!

Mattias-:
fixes #241

Label names does not follow the same matching rules as metric names.
https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -497,6 +497,11 @@ public class CloudWatchCollector extends Collector {
       return s.replaceAll("[^a-zA-Z0-9:_]", "_").replaceAll("__+", "_");
     }
 
+    private String safeLabelName(String s) {
+      // Change invalid chars to underscore, and merge underscores.
+      return s.replaceAll("[^a-zA-Z0-9_]", "_").replaceAll("__+", "_");
+    }
+
     private String help(MetricRule rule, String unit, String statistic) {
       if (rule.help != null) {
           return rule.help;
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #250, opened 2020-01-20
correct config.yml steps for building own image

deltamualpha:
The ONBUILD command expects `config.yml` to be in the same directory as the `Dockerfile`, not nested in a config directory. The readme was unclear on this point, causing me around five minutes of confusion until I read the parent `Dockerfile`. (This is a separate fix from #204.)


deltamualpha:
(and fixed the signoff.)

brian-brazil:
Thanks!

Issue #72, opened 2017-09-03
add a CMD instruction for the dockerfile like the `prom/prometheus` or `prom/alertmanger` containers


ajeeshpu:

ajeeshpu:
while building child images its necessary to have the config.yml in context since the onbuild performs an ADD
it will be nice to have a CMD option like the `prom/prometheus` or `prom/alertmanger` containers

brian-brazil:
I believe #254 resolved this.

PR #254, opened 2020-02-27
Remove onbuild

steve-exley:
The ONBUILD statement in the Dockerfile makes a assumptions about why the image may be built which are very specific and creates unecessary  future dependencies.  It's not really what the 'ONBUILD' command [was introduced for](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/). Removing it better aligns with the Docker Principle of build once, use anywhere and not having dependencies on the OS.

The intended capability of adding a `config.yml` file into the image can still easily be achieved.

https://github.com/prometheus/cloudwatch_exporter/issues/72 seems to be saying the same thing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,6 @@ EXPOSE 9106
 
 WORKDIR /
 RUN mkdir /config
-ONBUILD ADD config.yml /config/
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

PR #259, opened 2020-03-10
Implement describe() to prevent the scrape on startup

louisfelix:
Hi @brian-brazil ,

When we first used the Cloudwatch exporter, we noticed that the exporter was doing an initial scrape on startup. Although we didn't pay too much attention to it initially, this behaviour now slows down a lot our automated deployments, especially on large setups where multiple exporters are called and where scrape durations are around 30 seconds.

This PR will makes the collector to implement `Describable` in order to prevent the `CollectorRegistry` to fall back on the initial `collect()` on registration.

Thanks!


brian-brazil:
Thanks!

PR #275, opened 2020-05-15
allow maven-compiler-plugin to compile without a JDK downgrade

rothsa:
**Before:**
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] Source option 6 is no longer supported. Use 7 or later.
[ERROR] Target option 6 is no longer supported. Use 7 or later.
[INFO] 2 errors
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
**After:**
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------

Alternatively, one could also remove this so that the default takes over: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,8 +113,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
       <!-- Build a full jar with dependencies -->


brian-brazil:
What JDK are you trying to compile with that it fails?

rothsa:
@brian-brazil If I understand the documentation that I linked in this description correctly, -source and -target refer to the JVM to be compiled with. So, that would be 1.6. and 1.8 if this pull is merged. The withdrawal of plugin support from the earlier JVMs, I do not have a reference for, except for the ERROR that you see above.

rothsa:
$ java --version
openjdk 14.0.1 2020-04-14
OpenJDK Runtime Environment AdoptOpenJDK (build 14.0.1+7)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 14.0.1+7, mixed mode, sharing)

brian-brazil:
Thanks!

Looks like our docs say Java 8 already.

Issue #134, opened 2018-08-22
Reduce number of api operations by using GetMetricData API instead of GetMetricStatistics API


prasannamahajan:
You can request 100 metrics using GetMetricData and 1 metric using GetMetricStatistics.

Pricing of GetMetricData and GetMetricStatistics is 
$0.01/1,000 metrics requested using GetMetricData
$0.01/1,000 GetMetricStatistics

So if you use GetMetricData, the overall cost would become 1% of current cost. (huge saving)

brian-brazil:
That sounds promising, though I think it'd be less than that if you were requesting multiple statistics.

Would this also work for things like billing that only post stats once a day?

prasannamahajan:
whatever woks with GetMetricStatistics will work with GetMetricData.
If you request multiple statistics of same metric in single GetMetricData then it would be counted as 1 request.

ikben:
This may also allow people to use metric math (although this has probably impact on the configuration syntax), allowing for more useful data from cloudwatch:
Eg;
Getting (average) requests per seconds by dividing the per-minute metric by 60

richardTowers:
Sadly the pricing of `GetMetricData` and `GetMetricStatistics` doesn't work like that - 

```
$0.01/1000 metrics requested using GetMetricData
...
$0.01/1,000 GetMetricStatistics ... requests
```

So although you can make 100 times fewer calls (requesting 100 metrics in each), the `GetMetricData` is billed _per metric_ whereas the `GetMetricStatistics` call is billed _per request_, so it doesn't actually change the cost. (I checked this interpretation of AWS' pricing model with AWS support).

The place where the pricing does differ is in the number of datapoints returned _for a single metric_ - `GetMetricData` supports far more datapoints, so if you can poll less often and request a lot of data in each poll then you can save costs that way. Sadly that adds latency that you probably don't want.

brian-brazil:
Would using GetMetricData be an option for reducing the number of RPC calls and thus how long it takes to gather data?

frioux:
I think it would reduce the number of calls and time, if I am understanding correctly.  I could try to implement a patch if you are interested

brian-brazil:
Please go ahead. We'll only find out by trying it.

frioux:
[...]

PR #296, opened 2020-08-13
Use `RecentlyActive` feature to only list metrics older than 3 hours when necessary

nathan-beneke:
I have the problem that a lot of the dimensions on my CloudWatch metrics change frequently. Because the default behaviour of `listMetrics()` is to return every metric that matches your query and has been updated in the last 2 weeks, this causes the scrape time to be very high. With this option, we can limit the returned metrics to only those which have been updated in the last 3 hours. I also had to update the AWS SDK to allow this, as the version currently being used doesn't have this feature

cc @brian-brazil


brian-brazil:
If we're using this it shouldn't be configurable, we already know the range_seconds in use.

nathan-beneke:
So you would suggest to set this option if `range_seconds` < 3 hours? I guess that also makes sense

brian-brazil:
Yes.

nathan-beneke:
@brian-brazil I agree that's a lot simpler actually. Changed it to do that and updated the test to make sure that it's actually still possible to get metrics older than 3 hours with this setup

nathan-beneke:
I see this was also talked about in this open issue but never pursued further. With regard to the concern with S3 and similar metrics, I think it's okay because for those metrics the `range_seconds` has to be very high anyways https://github.com/prometheus/cloudwatch_exporter/issues/134#issuecomment-671185446_

brian-brazil:
Thanks!

guusvw:
Hey @brian-brazil 
What is your policy on building new releases? ;)

brian-brazil:
As-needed basically, I've added it to my todo list.

kjpianeta:
Hi @brian-brazil should we expect a new release with this in the next week(s), month(s)? Thanks.

PR #314, opened 2020-12-16
Pull latest openjdk:11 base image when building the Docker image

LS80:
This ensures that when a new Docker image is released it will be based on the latest available OpenJDK 11 image.

The currently released `0.9.0` Docker image is based on OpenJDK 11.0.8 which has some known vulnerabilities.

As the Dockerfile specifies a non-specific base image tag the image will be built with whatever version of OpenJDK 11 happens to be cached locally. With this change it will force the latest image pointed to by the tag `11-jre-slim` to be used as the base image, currently `11.0.9.1-jre-slim` (`sha256:28b59dc9a129c75349418e3b75508fd8eef3b33dd7d796079d1d19445d907776`).

The alternative to this would be to use a more specific tag for the base image, so `11.0.9-jre-slim` maybe.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - run:
           command: docker run --privileged linuxkit/binfmt:v0.7
       - run: |
-          docker build -t quay.io/prometheus/cloudwatch-exporter:latest .
+          docker build -t quay.io/prometheus/cloudwatch-exporter:latest --pull .
           if [[ -n "$CIRCLE_TAG" ]]; then
             docker tag -t quay.io/prometheus/cloudwatch-exporter:"${CIRCLE_TAG}" quay.io/prometheus/cloudwatch-exporter:latest
           fi


.circleci/config.yml
```diff
@@ -19,25 +19,25 @@ jobs:
 
   publish_image_to_quay:
     docker:
-      - image: circleci/openjdk:8-jdk-stretch
+      - image: docker.io/docker:20.10-git
```
brian-brazil, line 22: I don't think we should be adding dependencies on docker hub given the rate limits.

.circleci/config.yml
```diff
@@ -19,25 +19,25 @@ jobs:
...
-          docker build -t quay.io/prometheus/cloudwatch-exporter:latest .
+          docker build -t quay.io/prometheus/cloudwatch-exporter:latest --pull .
           if [[ -n "$CIRCLE_TAG" ]]; then
             docker tag -t quay.io/prometheus/cloudwatch-exporter:"${CIRCLE_TAG}" quay.io/prometheus/cloudwatch-exporter:latest
           fi
       - run: docker images
       - run: docker login -u $QUAY_LOGIN -p $QUAY_PASSWORD quay.io
       - run: |
-          docker push quay.io/prometheus/cloudwatch-exporter:latest
+          docker push --quiet quay.io/prometheus/cloudwatch-exporter:latest
```
brian-brazil, line 38: I don't think hiding output makes sense here.

brian-brazil (commented):
I don't understand this change, there seem to be several unrelated things going on.
Can you explain the motivation behind it, and what problem(s) you're trying to solve?

.circleci/config.yml
LS80: True. Although I doubt it would be a problem.

.circleci/config.yml
LS80: On CircleCI the carriage returns for the push progress don't work so you can end up with a lot of log spam. I'll drop the commit.

brian-brazil:
Thanks!

PR #317, opened 2020-12-29
Add logging implementation for AWS SDK

LS80:
Fixes the error `Failed to load class "org.slf4j.impl.StaticLoggerBinder"`

See [AWS docs](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/logging-slf4j.html) for more information.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,11 @@
       <artifactId>resourcegroupstaggingapi</artifactId>
       <version>2.15.20</version>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>1.7.30</version>
+    </dependency>
     <!--
       Older versions of jackson-databind have remote code execution vulnerabilities
       CVEs - 2018-5968, 2017-17485, 2017-7525, 2017-15095, 2018-7489, 2018-19360,


brian-brazil:
Thanks, those docs don't mention this pom entry though. Are you sure this is right?

LS80:
As I understand it this tells SLF4J to use the standard Java Logging Framework. This example is not explicitly mentioned in the AWS docs but it seems to work. I tried the Apache Log4j example in the AWS docs but it seems it doesn't play nicely with Java 11.

There's some useful information in the [SLF4J docs](http://www.slf4j.org/manual.html):

> To switch logging frameworks, just replace slf4j bindings on your class path. For example, to switch from java.util.logging to log4j, just replace slf4j-jdk14-1.7.28.jar with slf4j-log4j12-1.7.28.jar.

> SLF4J does not rely on any special class loader machinery. In fact, each SLF4J binding is hardwired at compile time to use one and only one specific logging framework. For example, the slf4j-log4j12-1.7.28.jar binding is bound at compile time to use log4j. In your code, in addition to slf4j-api-1.7.28.jar, you simply drop one and only one binding of your choice onto the appropriate class path location. Do not place more than one binding on your class path. Here is a graphical illustration of the general idea.

image

brian-brazil:
Thanks!

PR #318, opened 2020-12-29
Use non-root user for Docker image

LS80:
Running containers as a non-root user is considered security best practice.

Replaces the `MAINTAINER` instruction (which is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)) with a label.

Excludes the `target` directory from the Docker build context. This is only useful if building the image locally after running `mvn package`.


brian-brazil:
Might this cause file permissions issues for existing users?

LS80:
It's possible. In my case I just needed to add the following to an EKS pod spec for the AWS IAM roles for service accounts to work.

```yaml
      securityContext:
        fsGroup: 1000
```

Any sidecar containers may need to run as the same user if there is a shared volume that it writes to

```yaml
         securityContext:
           runAsUser: 1000
```

brian-brazil:
We've seen such things cause problems for users elsewhere, and have previously had to revert changes like this due to it.

PR #323, opened 2021-01-29
Use a copy constructor instead of Cloneable

matthiasr:
SonarLint strongly advises against clone(). As advised, use a copy
constructor instead. For now, keep it simple and do not attempt a
deep-copy.

Signed-off-by: Matthias Rampke <matthias@prometheus.io>


matthiasr:
ah, meant to add a link there: https://www.artima.com/intv/bloch.html#part13

PR #345, opened 2021-05-18
Update awssdk to 2.16.64

srhaber:
This makes it possible to query RDS Proxy metrics. https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy.html#rds-proxy.monitoring


matthiasr:
Thank you!

Issue #237, opened 2019-12-03
Exporter not picking up AWS/Kafka metrics 


jim5252:
I am able to get metrics using the exporter tool for other AWS resources but when attempting to use for Kafka, no metrics are being retrieved:

`HELP tagging_api_requests_total API requests made to the Resource Groups Tagging API
 TYPE tagging_api_requests_total counter

 HELP cloudwatch_exporter_scrape_duration_seconds Time this CloudWatch scrape took, in seconds.

 TYPE cloudwatch_exporter_scrape_duration_seconds gauge
cloudwatch_exporter_scrape_duration_seconds 0.0403737

 HELP cloudwatch_exporter_scrape_error Non-zero if this scrape failed.

 TYPE cloudwatch_exporter_scrape_error gauge

cloudwatch_exporter_scrape_error 0.0

 HELP cloudwatch_requests_total API requests made to CloudWatch

 TYPE cloudwatch_requests_total counter

cloudwatch_requests_total{action="listMetrics",namespace="AWS/Kafka",} 16.0`

There is not any examples of using Kafka in the repository, does anyone know if it has been successfully used for Kafka before?

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

jim5252:
Thank you, i shall direct my question there.

RmMsr:
Hi @jim5252 ,

we experienced the same problem. After quite a while of experimenting we noticed that the naming is crucial. For example the dimensions must be identical to the AWS naming with case and whitespace.

Example:

```
metrics:
  - aws_namespace: AWS/Kafka
    aws_metric_name: KafkaDataLogsDiskUsed
    aws_dimensions: [Broker ID, Cluster Name]
    aws_statistics: [Maximum]
```

vasrem:
BTW guys, JFYI you have to specify **both** aws_dimensions to get the metrics back as @RmMsr described.

pmccau:
Note that to get any tag filters to work you need to pass in a comma-separated **string** value in lieu of a list (for some reason?), for example:

```yaml
       aws_dimensions: [Broker ID, Cluster Name]
       ...
       aws_tag_select:
          tag_selections:
            owner: ["your_owner_here"]
[...]

PR #350, opened 2021-06-25
Update aws_dimensions config documentation

agarella:
I ran into an issue where I could not find the right configuration to retrieve Kafka metrics from Cloudwatch.

Then I saw this comment: https://github.com/prometheus/cloudwatch_exporter/issues/237#issuecomment-581895552, and found out that in order for the metrics to work, you need to specify exactly all the aws_dimensions in order to be able to retrieve any metrics at all.

This change in the README reflects that.

However, the way it currently works I would suggest removing the `aws_dimensions` option altogether, as it only causes confusion without providing any actual configuration options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ role_arn   | Optional. The AWS role to assume. Useful for retrieving cross accou
 metrics  | Required. A list of CloudWatch metrics to retrieve and export
 aws_namespace  | Required. Namespace of the CloudWatch metric.
 aws_metric_name  | Required. Metric name of the CloudWatch metric.
-aws_dimensions | Optional. Which dimension to fan out over.
+aws_dimensions | Required. This should contain exactly all the dimensions available for a metric. Run `aws cloudwatch list-metrics` to find out which dimensions you need to include for your metric.
 aws_dimension_select | Optional. Which dimension values to filter. Specify a map from the dimension name to a list of values to select from that dimension.
 aws_dimension_select_regex | Optional. Which dimension values to filter on with a regular expression. Specify a map from the dimension name to a list of regexes that will be applied to select from that dimension.
 aws_tag_select | Optional. A tag configuration to filter on, based on mapping from the tagged resource ID to a CloudWatch dimension.  


matthiasr:
Thank you! I created #351 for the follow up.

README.md
```diff
@@ -64,7 +64,7 @@ role_arn   | Optional. The AWS role to assume. Useful for retrieving cross accou
 metrics  | Required. A list of CloudWatch metrics to retrieve and export
 aws_namespace  | Required. Namespace of the CloudWatch metric.
 aws_metric_name  | Required. Metric name of the CloudWatch metric.
-aws_dimensions | Optional. Which dimension to fan out over.
+aws_dimensions | Required. This should contain exactly all the dimensions available for a metric. Run `aws cloudwatch list-metrics` to find out which dimensions you need to include for your metric.
```
brian-brazil, line 67: This is incorrect, the user must chose which ones they want and this can be empty.

agarella: Are you sure? I couldn't get it to work with this example config:

```
config: |-
  region: us-west-2
  metrics:
   - aws_namespace: AWS/Kafka
     aws_metric_name: SumOffsetLag
```

What is the expected behavior when `aws_dimensions` is empty?

Only after adding these exact dimensions:
```
aws_dimensions: [Consumer Group, Cluster Name, Topic]
```

was I able to retrieve any metrics. Any other combination did not work for me.

brian-brazil: Yes, I'm sure. When it's empty, you should get a fully aggregated result. Unfortunately Cloudwatch at times doesn't follow its own data model for certain services, and will instead return nothing for certain combinations of dimensions.

agarella: Well that's unfortunate then because it does not work for the metric that I mentioned. Which means that the `aws_dimensions` is not optional in all cases.

While my update to the docs may not be technically correct, it does suggest a configuration that is guaranteed to work.

I would suggest adding to the documentation that `aws_dimensions` is not optional for all metrics. With the addition to try adding all the dimensions from `aws cloudwatch list-metrics` in case nothing is returned. That would help future users like myself save a lot of time and trial & error while setting this up.

brian-brazil: > While my update to the docs may not be technically correct, it does suggest a configuration that is guaranteed to work.

I'd not presume that unfortunately.

matthiasr: *shakes fist at computers* thank you @brian-brazil for setting us straight

@agarella I agree that we should document this fully and help users as much as possible. Do you want to take a stab at this? I think it deserves a whole section with a header so that it stands out.

Issue #357, opened 2021-08-18
Add AWS Account ID to labels


sean-nixon:
We have a multi-account environment and would like to use the `role_arn` configuration to export metrics across our accounts from one instance of the cloudwatch_exporter. However, there seems to be no way to differentiate between metrics from the different accounts. Because of this, role_arn doesn't really help us and we'd be forced to run one instance of the exporter for each account, pull from each of them, and do some intricate relabeling in the prometheus configuration based on the metalabels from the scrape targets.

It would be highly useful if the source AWS account ID was added as a label (perhaps `account_id`) to each of the exported metrics. This could be determined from a GetCallerIdentity to STS. If a role_arn is provided, it could also be parsed from that value.

Barring this functionality, an ability to specify custom labels in each metric would also suffice.

Would the above functionality be considered in this project? If so, I would be willing to give a go at a PR.

matthiasr:
Can you give an example how you would configure the exporter to pull metrics from multiple accounts with one configuration?

matthiasr:
I am leaning towards

> Barring this functionality, an ability to specify custom labels in each metric would also suffice.

since this would be more general. Someone else may not want to use account IDs but some other identifier, but they'd know what they want when writing the configuration.

sean-nixon:
@matthiasr On second look at the documentation, it looks like I was probably mistaken and it's not possible to pull metrics from multiple accounts with one configuration since `role_arn` is a global configuration. I was initially thinking a configuration like this, but I believe it's invalid now.

```
region: us-west-2
metrics:
 - aws_namespace: AWS/ELB
   aws_metric_name: RequestCount
   aws_dimensions: [AvailabilityZone, LoadBalancerName]
   aws_dimension_select:
     LoadBalancerName: [myLB]
   aws_statistics: [Sum]
   role_arn: <account1>
 - aws_namespace: AWS/ELB
   aws_metric_name: RequestCount
   aws_dimensions: [AvailabilityZone, LoadBalancerName]
   aws_dimension_select:
     LoadBalancerName: [myLB]
   aws_statistics: [Sum]
   role_arn: <account2>
```

I think the request still has value since it removes the need to do relabeling in the prometheus scraper configuration. I understand your logic for favoring custom labels over only the account ID and that would be great to have.

matthiasr:
> it removes the need to do relabeling in the prometheus scraper configuration

[...]

PR #366, opened 2021-10-20
Add the possibility to add custom labels to metrics

jebeaudet:
My use case is to have multiple metrics with the same `aws_metric_name` but different `period_seconds` and/or `range_seconds`. Example : 
```
metrics:
    - aws_dimension_select_regex:
        DBInstanceIdentifier:
        - some-identifier*
      aws_dimensions:
      - DBInstanceIdentifier
      aws_metric_name: CPUUtilization
      aws_namespace: AWS/RDS
      aws_statistics:
      - Average
      delay_seconds: 90
    - aws_dimension_select_regex:
        DBInstanceIdentifier:
        - some-identifier*
      aws_dimensions:
      - DBInstanceIdentifier
      aws_metric_name: CPUUtilization
      aws_namespace: AWS/RDS
      aws_statistics:
      - Average
      delay_seconds: 90
      period_seconds: 604800
      range_seconds: 604800
```
With the current setup, this will produce 2 metrics with the exact same labels but different values : 
```
# HELP aws_rds_cpuutilization_average CloudWatch metric AWS/RDS CPUUtilization Dimensions: [DBInstanceIdentifier] Statistic: Average Unit: Percent
# TYPE aws_rds_cpuutilization_average gauge
aws_rds_cpuutilization_average{job="aws_rds",instance="",dbinstance_identifier="someidentifier-1",} 9.508491808196803 1634693160000
# HELP aws_rds_cpuutilization_average CloudWatch metric AWS/RDS CPUUtilization Dimensions: [DBInstanceIdentifier] Statistic: Average Unit: Percent
# TYPE aws_rds_cpuutilization_average gauge
aws_rds_cpuutilization_average{job="aws_rds",instance="",dbinstance_identifier="someidentifier-1",} 9.092518065733273 1634088420000
```
Which is useless in prometheus. With this PR, we'll be able to distinguish between the two using labels that you set in the config. I thought about just setting the `delay_seconds`, `period_seconds` and `range_seconds` as default labels however I thought that having specific custom labels (something like `period="weekly"` in this case) is clearer. I'm open to changing my PR for this if you prefer though, it's a matter of preference I believe in the end.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ metrics:
    aws_dimension_select:
      LoadBalancerName: [myLB]
    aws_statistics: [Sum]
+   additional_labels: 
+     lb_type: public
 ```
 
 A similar example with common options and `aws_tag_select`:
@@ -77,13 +79,14 @@ delay_seconds | Optional. The newest data to request. Used to avoid collecting d
 range_seconds | Optional. How far back to request data for. Useful for cases such as Billing metrics that are only set every few hours. Defaults to 600s. Can be set globally and per metric.
 period_seconds | Optional. [Period](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/cloudwatch_concepts.html#CloudWatchPeriods) to request the metric for. Only the most recent data point is used. Defaults to 60s. Can be set globally and per metric.
 set_timestamp | Optional. Boolean for whether to set the Prometheus metric timestamp as the original Cloudwatch timestamp. For some metrics which are updated very infrequently (such as S3/BucketSize), Prometheus may refuse to scrape them if this is set to true (see #100). Defaults to true. Can be set globally and per metric.
+additional_labels | Optional. Map of additional labels to be set on the metric. Can be used to distinguish identical `aws_metric_name` with different `range_seconds` for example. Can only be set per metric.
 
 The above config will export time series such as
 ```
 # HELP aws_elb_request_count_sum CloudWatch metric AWS/ELB RequestCount Dimensions: ["AvailabilityZone","LoadBalancerName"] Statistic: Sum Unit: Count
 # TYPE aws_elb_request_count_sum gauge
-aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="mylb",availability_zone="eu-west-1c",} 42.0
-aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="myotherlb",availability_zone="eu-west-1c",} 7.0
+aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="mylb",availability_zone="eu-west-1c",lb_type="public",} 42.0
+aws_elb_request_count_sum{job="aws_elb",instance="",load_balancer_name="myotherlb",availability_zone="eu-west-1c",lb_type="public",} 7.0
 ```
 
 If the `aws_tag_select` feature was used, an additional information metric will be exported for each AWS tagged resource matched by the resource type selection and tag selection (if specified), such as
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -70,6 +70,7 @@ public class CloudWatchCollector extends Collector implements Describable {
       int periodSeconds;
       int rangeSeconds;
       int delaySeconds;
+      Map<String,String> additionalLabels;
       List<Statistic> awsStatistics;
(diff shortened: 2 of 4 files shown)


matthiasr:
Hmm, I am very unsure about this. Generally, "add arbitrary labels" is not something exporters should support, this is what relabeling is for … on the other hand, relabeling won't help in this particular case.

I would like to take a step back though: why do you need multiple iterations of the same metric? What prevents having only the shorter period, using PromQL queries to calculate the longer aggregations?

jebeaudet:
Like you said, relabeling is useless here since it's the exact same metric and I have no way to distinguish both metrics. I'd be happy to do it on the prom side but I can't, hence this PR. 

The problem in my case is regarding cost and scalability of our prometheus infrastructure. At the scale my company is at, it's hard to buid a relatively cheap and reliable infrastructure to store the thousands of metrics it's being sent and for a long period of time. The direction we took was to delete stored metrics after 24h. This works for immediate monitoring of the type "the cpu is 100% right now" but for long term trend monitoring, it's not possible.

Therefore, a sane alternative to this is to delegate to cloudwatch this and query it to export the metric over a longer period of time.

Like I mentioned in the OP, we could also export the `delay_seconds`, `period_seconds` and `range_seconds` as default labels which would also make this possible and would not be an "arbitrary labels" solution which you don't seem to like. 

Thanks!

matthiasr:
I'm going to have to decline this request. Exporting the same metric multiple times is not something I want to support in the container, because of the complexity it brings. You can get the same effect by running multiple exporter instances with the different configurations; this avoids conflicts within the exporter and lets you use relabeling.

This workaround is more flexible because it allows different configurations and labeling schemes in general, not just for this particular use case. The exporter is lightweight enough that this will not cause excessive resource usage compared to the Prometheus collecting the metric.

jebeaudet:
It's your project, but I really don't understand your justification. How is running 2 instances of the exporter more flexible than just having a single instance exporting all the proper metrics with correct labels? You'll have to name your pods properly to say hey relabel metrics from pod "cloudwatch-exporter-daily" and add a label "daily" to all its metrics. It somewhat couples your prom configuration to your deployment and to me it just looks inelegant.

I also fail to see the complexity this PR brings. The exporter already exports the same metric multiple times, it just does it without the proper labels so it's useless. If you really have this requirement, it's missing a uniqueness validation in the code then.

This being said, thanks for taking the time to review the PR, we'll just continue with our fork instead.

matthiasr:
> It somewhat couples your prom configuration to your deployment and to me it just looks inelegant.

I think this is where we value different things. To me (and in the approach of the Prometheus project generally), adding variation and customization to exporters is "inelegant", and moving the smarts into the scraping configuration is the right way™. I completely agree that it could be the other way around, but this is the path we chose.

In other exporters, such as the [mysqld_exporter](https://github.com/prometheus/mysqld_exporter#filtering-enabled-collectors], [collectd_exporter](https://github.com/prometheus/node_exporter#filtering-enabled-collectors), and [blackbox exporter](https://github.com/prometheus/blackbox_exporter#prometheus-configuration), we avoid having to deploy multiple exporter instances by accepting request parameters. What parameters would be needed here so that you can have one exporter, with one set of configuration, that serves the different use cases?

> The exporter already exports the same metric multiple times, it just does it without the proper labels so it's useless. If you really have this requirement, it's missing a uniqueness validation in the code then.

Indeed! I created #383 for that.

matthiasr:
What I mean with "more flexible" is that it does not only solve your use case, but also for example [scraping multiple accounts](https://github.com/prometheus/cloudwatch_exporter/issues/357), and scraping certain metrics more (less) frequently.

jebeaudet:
Thanks for explaining the reasoning. But I still can't say I agree, to me the "smarts" in what you propose lay in me deploying 2 different exporters, not in the now possible relabeling possibilities that having 2 exporters creates. And having to deploy 2 exporters just to export the same metric twice, along with some relabeling rules, is so much more work than just exposing a configurable extra set of labels, or just directly `period_seconds` which is already part of the configuration. 

Thanks for the linked issue, it is similar indeed (with a similar ending 😂) but to me the difference is that he can do it by relabeling, I can't (unless I deploy 2 instances).

I've looked at your linked repo and AFAICT they all use request parameters to filter out specific metrics. They don't use the parameters to scrape a different metric on demand. It would also require a bigger rework because `io.prometheus.client.exporter.MetricsServlet` also support only filtering by metric name, and this will not work if you do #383 because I won't be able to scrape the same metric twice. 

P.S. Unrelated to this issue but you're leaking an instance of `FileReader` in `WebServer` which this PR fixed.

PR #374, opened 2021-11-19
circleci: Pass BUILDX_PLATFORMs when building multi-arch images

lht:
Related to #363. Fixes #337.

PR #372, opened 2021-11-15
Update Java/client_java

SuperQ:
* Update to openjdk 17.
* Update client_java to 0.12.0.

Signed-off-by: SuperQ <superq@gmail.com>

PR #363, opened 2021-09-27
update circleci to build multi-arch images

garrygerber:
using "docker buildx" now building for arch's linux/amd64 and linux/arm64
related to issue #337

PR #376, opened 2021-11-19
Fix Docker builds on linux/arm64

matthiasr:
Follow-up to #372, #374, #363.

Update the base images used in docker builds. Install Maven from the
tarball, apt pulls in the wrong JDK as a dependency.

Set up the context and builder for buildx.

Work around a bug in Java 13+ on linux/arm64 by using the older "vfork"
method. Update the surefire plugin.

Always run the image build, even for PRs, so that we can catch breakage
here before it hits the main branch.

cc @lht


SuperQ (approved):
LGTM, nice.

PR #386, opened 2021-12-06
feat: use jlink rather than jdk

greut:
It makes the final image smaller.

- 221MB, https://hub.docker.com/layers/prom/cloudwatch-exporter/latest/images/sha256-6a5afd6d7f1cbc90bfea77f0b92942d57422f9a91d441bb293837693a388e867?context=explore
- 71MB, https://hub.docker.com/layers/greut/cloudwatch_exporter/latest/images/sha256-b455dea048b56ad3ce98b2a43669b42a89df9c09aab0099355283d84790180b0?context=explore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,6 +149,5 @@ workflows:
               only: /.*/
             branches:
               only: master
-
 orbs:
   prometheus: prometheus/prometheus@0.14.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,8 @@ jobs:
     working_directory: ~/circleci-java
     docker:
       - image: circleci/openjdk:17-jdk-buster
+    environment:
+      MAVEN_OPTS: "-Djdk.lang.Process.launchMechanism=vfork"
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


matthiasr:
Interesting, I like the idea especially since it promises reduced memory usage as well. However, I am hesitant to change the Docker image in an incompatible way, especially since I just updated the JDK and base image. What do you think about splitting this out into a separate `-minimal` set of images? Would that be possible in one go or would it require building from scratch twice?

Would it be beneficial to generate a custom launcher script and use it as the entry point, or is that unnecessary?

matthiasr:
Totally fine to keep out of scope, but it slightly bugs me that we are building the exporter twice for amd64 and arm64. Using the same jars in both images would cut down on the build time significantly, especially if that build can happen on the native host architecture.

greut:
All the _magic_ done at the end of the upstream Dockerfile is probably useful as well. https://github.com/docker-library/openjdk/blob/41610205127c2a4f0af64f68b859726993503a89/17/jdk/slim-bullseye/Dockerfile#L67-L86

matthiasr:
@greut do you still want to work on this?

greut:
By splitting that in a separate image, for example?

matthiasr:
Yes!

PR #388, opened 2021-12-08
Improve health and ready endpoints

ssboisen:
This PR changes the health endpoints so they reflect the latest status of the collector.

Recently during an AWS outage our cloudwatch metrics exporter (running in kubernetes) broke because the outage caused permissions to invalidate which meant the collector could no longer access the AWS api. 

The solution was to manually delete the pod however it would have been great if the liveness endpoint had detected the problem so it could have resolved the situation automatically. This PR should help with that.


matthiasr:
I am not sure about this. I don't think any service should start crashing or failing liveness probes because some dependency is unavailable temporarily.

When CloudWatch is broken, what does a restart improve? When CloudWatch recovers, the exporter should start getting metrics again, if it does not that's a bug – restarting through a liveness probe may work around that but it's not a solution.

matthiasr:
Thank you for the idea, but I decided to close this. If there is an issue that the exporter does not recover from by itself, we should fix that, rather than requiring restarts. And we should never force a restart only because CloudWatch had a hickup.

PR #395, opened 2021-12-30
remove redundancy in examples of EBS & NetworkELB

arnoldyahad:
Hey guys,
While working with your examples I noticed some metrics repeat themselves (either exactly with the same statistic or with a different statistic - which can be merged into one item)


matthiasr (approved):
Thank you!

PR #398, opened 2022-02-02
Add JVM hotspot metrics

mindw:
Should be useful for sizing and resource use cases.


matthiasr (approved):
Nice and simple, I like it 😄

PR #401, opened 2022-02-25
feat: migrate from OpenJDK to Eclipse Temurin

greut:
Since #386, they started to provide a JRE image. As `openjdk:17-slim-bullseye` is `openjdk:17-jdk-slim-bullseye`; it really creates a big image for little benefits.

Eclipse Temurin is the new cool kid, http://whichjdk.com/ - https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/


matthiasr:
Very neat, thank you!

PR #403, opened 2022-03-04
fix: sha256sum not working as expected

greut:
Sorry for the aborted release...


greut:
Done by #401 already.

PR #418, opened 2022-04-26
Added automatic code formatter

### LINKED ISSUE #417: Unify and enforce formatting
This came up in [#414](https://github.com/prometheus/cloudwatch_exporter/pull/414#issuecomment-1108298222) and I ran into it a few times.

Unified and automatically formatable styling makes it easier for different contributors to work on the code. Instead of piece-meal changes that make individual PRs difficult to follow, I would like to do one big bang reformat, ideally while there are no other open PRs.

What are the most common formatters? Ideally we should not require any special setup from users with auto-formatting IDEs. At the least, users of (say) IntelliJ, Eclipse, and VS Code don't need to do anything to stop their editor from enforcing the wrong format; better would be if these also automatically enforce the _right_ format.

In any case, we should enforce this formatting in PR pipelines, probably in a separate job for clarity.

or-shachar:
Resolves #417 :-) 
Note: this requires JDK > 11

The first commit is adding a maven plugin that can check and auto-fix formatting + adds CircleCI step that fails if somehow the developer forgot to format. You can see the CircleCI failure (as the code is not formatted).

The second commit is after applying the plugin format.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,16 @@ jobs:
           paths:
             - ~/.m2
           key: circleci-demo-java-spring-{{ checksum "pom.xml" }}
+      - run: 
+          name: Check formatting
+          command: |
+            if (mvn fmt:check);
+            then
+              echo "Code is well formatted"
+            else
+              echo "Code is not formatted well. Please run 'mvn fmt:format' and commit changes"
+              exit 1
+            fi 
       - run: mvn package
   publish_image:
     docker:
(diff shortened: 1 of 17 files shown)


matthiasr:
I like it!

I think JDK 11 is totally fine, we don't test on anything but 17 anyway, and since this isn't a library I am not worried about broad legacy support.

matthiasr:
#417 is merged, could you rebase this PR so we can get it merged, or allow me to make modifications?

matthiasr:
Beautiful!

PR #421, opened 2022-05-07
Update maven dependencies (jetty 10->11)

### LINKED ISSUE #390: Update to Jetty 11
Follow-up to #389: update to Jetty 11, and follow the javax -> jakarta transition.

or-shachar:
Fixes #390 ... + 
1. Moving repeated versions to maven properties (aws, prometheus)
2. Updating prometheus - 0.14.1 --> 0.15.0
3. Updating jetty - 10.0.8 --> 11.0.9
4. commons-coden 1.13 --> 1.15
5. Updating maven compiler plugin and release level.
6. Updated CircleCI image to `cimg/openjdk:17.0.3` (instead of deprecated openjdk image)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/circleci-java
     docker:
-      - image: circleci/openjdk:17-jdk-buster
+      - image: cimg/openjdk:17.0
     steps:
       - checkout
       - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/circleci-java
     docker:
-      - image: circleci/openjdk:17-jdk-buster
+      - image: cimg/openjdk:17.0.3
```
SuperQ, line 6: Can we pin this to `cimg/openjdk:17.0` so we only lock the major/minor version?

SuperQ (commented):
Thanks, looks mostly good. Just want to change the version pinning to be less strict.

or-shachar:
Thanks for the CR @SuperQ ! 🙏

SuperQ (approved):

matthiasr (approved):
Thank you both!

PR #423, opened 2022-05-08
Fixed null unit in HELP in case of GetMetricStatistics

### LINKED ISSUE #422: "Unit" is missing from HELP after #414
Minor regression that was probably caused by my #414. The unit is missing from the HELP:
### before:
```
# HELP aws_sqs_sent_message_size_average CloudWatch metric AWS/SQS SentMessageSize Dimensions: [QueueName] Statistic: Average Unit: Bytes
```

### after
```
# HELP aws_sqs_sent_message_size_average CloudWatch metric AWS/SQS SentMessageSize Dimensions: [QueueName] Statistic: Average Unit: null
```

I can probably fix it in case of `use_get_metric_data` is set to `false` but in case it is set to true - we don't really have a way to automatically set the unit as this is not part of the [GetMetricData response signature](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html).

|Quote from documentation|
|---|
|image |

Not sure if it's critical at all, but maybe we can add another optional field called "unit" and let the user fill this information manually.

or-shachar:
Partially fixes #422

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -652,7 +652,7 @@ public class CloudWatchCollector extends Collector implements Describable {
         if (values == null) {
           continue;
         }
-
+        unit = values.unit;
         List<String> labelNames = new ArrayList<>();
         List<String> labelValues = new ArrayList<>();
         labelNames.add("job");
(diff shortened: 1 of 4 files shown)


SuperQ (approved):
LGTM

matthiasr:
Thank you!

Issue #432, opened 2022-06-06
Incorrect Configuration?


mattystevenson:
Hi there,

Thanks for this! I'm trying to get it working to get data for Amazon GameLift which I think would be a welcomed alternative to CloudWatch. I've gotten things seemingly set up but I am not seeing any metrics. Here is what I get from the metrics screen for my CloudWatch Exporter. 

```
# HELP cloudwatch_requests_total API requests made to CloudWatch
# TYPE cloudwatch_requests_total counter
cloudwatch_requests_total{action="listMetrics",namespace="AWS/Redshift",} 611.0
cloudwatch_requests_total{action="listMetrics",namespace="AWS/ElastiCache",} 470.0
cloudwatch_requests_total{action="getMetricStatistics",namespace="AWS/GameLift",} 47.0
cloudwatch_requests_total{action="getMetricStatistics",namespace="AWS/S3",} 376.0
cloudwatch_requests_total{action="listMetrics",namespace="AWS/S3",} 94.0
cloudwatch_requests_total{action="listMetrics",namespace="AWS/GameLift",} 94.0
cloudwatch_requests_total{action="listMetrics",namespace="AWS/ELB",} 235.0
cloudwatch_requests_total{action="listMetrics",namespace="AWS/CloudFront",} 282.0
# HELP aws_s3_bucket_size_bytes_average CloudWatch metric AWS/S3 BucketSizeBytes Dimensions: [BucketName, StorageType] Statistic: Average Unit: Bytes
# TYPE aws_s3_bucket_size_bytes_average gauge
aws_s3_bucket_size_bytes_average{job="aws_s3",instance="",storage_type="StandardStorage",bucket_name="glsbe1stezmatt2",} 228661.0 1654314900000
aws_s3_bucket_size_bytes_average{job="aws_s3",instance="",storage_type="StandardStorage",bucket_name="glsbe1stezmatt",} 228660.0 1654314900000
aws_s3_bucket_size_bytes_average{job="aws_s3",instance="",storage_type="StandardStorage",bucket_name="glsbe1stezmatt1",} 228661.0 1654314900000
aws_s3_bucket_size_bytes_average{job="aws_s3",instance="",storage_type="StandardStorage",bucket_name="cloudtrail-awslogs-604060733806-d4tnpg08-isengard-do-not-delete",} 1.52625672E8 1654314900000
[...]

PR #436, opened 2022-06-09
Log warning in case of listDimensions returns empty list

or-shachar:
Relates to #432  - print warning in case dimensions is empty. 
Otherwise - it's hard to understand that configuration is wrong.

I wonder though if there is any scenario where the case of empty dimensions is actually okay. 
If there are - then this warning may be spamming and wrong.

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
+++ b/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
@@ -424,7 +424,6 @@ public class CloudWatchCollector extends Collector implements Describable {
         }
       }
     }
-
     return result;
   }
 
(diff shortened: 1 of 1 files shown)


matthiasr:
Hmm, good point. Googling around I found grafana/grafana#42798. Can we somehow distinguish between custom and AWS metrics?

or-shachar:
Nice catch! I removed the warning from [this case](https://github.com/or-shachar/cloudwatch_exporter/blob/fde6de56a473e18cabd4b2b4b2259d12946b0a2f/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java#L434).

IIUC the only way for it to work for a metric with no dimension is if we add an empty List to the dimensions list, so at the end of [this function](https://github.com/or-shachar/cloudwatch_exporter/blob/fde6de56a473e18cabd4b2b4b2259d12946b0a2f/src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java#L476), there must be at least one item in the returned list.

or-shachar:
Also maybe we need to document that this configuration with empty list actually supports metric with no dimensions (and that no value implicates metric with no dimension)
image

matthiasr:
Yes indeed, I think the documentation around dimensions could be less technical and more practical. Nonetheless, thank you for adding the warning!

PR #450, opened 2022-08-11
Disallow TRACE HTTP method

pharaujo:
Even though I believe there is no actual risk of [Cross-Site Tracing][xst] attacks in cloudwatch-exporter, disallowing
the `TRACE` HTTP method in Jetty reduces attack surface and makes security scans happier.

(Non-embedded) Jetty disables `TRACE` [in the default configuration][jetty-webapp] by forcing requests using that method be authenticated. I thought using a customizer would be cleaner and, as a bonus, would also be able to return a more accurate HTTP status code (405 Method Not Allowed).

[xst]: https://owasp.org/www-community/attacks/Cross_Site_Tracing
[jetty-webapp]: https://github.com/eclipse/jetty.project/blob/jetty-11.0.11/jetty-webapp/src/main/config/etc/webdefault.xml#L423-L440

### DIFF
--- a/src/main/java/io/prometheus/cloudwatch/DisallowHttpMethods.java
+++ b/src/main/java/io/prometheus/cloudwatch/DisallowHttpMethods.java
@@ -1,0 +1,25 @@
+package io.prometheus.cloudwatch;
+
+import java.util.EnumSet;
+import org.eclipse.jetty.http.HttpMethod;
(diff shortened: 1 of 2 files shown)


pharaujo:
@matthiasr can you have a look?

matthiasr:
Awesome, thank you!

PR #482, opened 2023-02-07
Fix measurement of cloudwatch_metrics_requested_total

epall:
As I read the existing code, it's trying to count the number of GetMetricData API calls made, but AWS is billing for every statistic within that. I'm seeing a discrepancy between what cloudwatch_exporter reports and what AWS bills us by a factor of 2-5x, depending on how I slice it.

Per the [AWS documentation](https://aws.amazon.com/cloudwatch/pricing/), "You can request up to five statistics for the same metric in a single GetMetricData API request. Additional statistics are billed as an additional metric."

So even though we can coalesce multiple statistics into a single API call, they're billed by AWS as if they were in individual API calls, so we should be measuring the full count, not dividing by five, to properly account for what AWS is charging.


matthiasr:
Indeed! FYI @or-shachar - this is a follow-up to #414

Issue #484, opened 2023-02-09
[validation config file ]:  command or other way to validate config file 


saharmethneni:
# Command or other way to validate the configuration file
### It will be better to validate the configuration file to detect errors before loading it .

This will allow the user to only load the correct configuration files and not break what is already running

matthiasr:
I agree, this would be a great thing to have. It could be a function of the regular exporter, if supplied with a flag it would check the configuration and exit (status code 0 if correct, 1 if incorrect).

As the base case, I think this should be a purely offline syntax check. However, given the nature of the exporter, it would be useful to have an extended / online check mode that actually connects to CloudWatch to validate that it can fetch the metrics, dimensions are correct, etc. Iit should not stop at the first error but rather list all (possible) problems.

PR #493, opened 2023-02-16
Use safe YAML loading

matthiasr:
We don't *need* unsafe YAML deserialization, so make sure we don't. It
is unlikely but not unthinkable that someone might load an untrusted
configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>1.32</version>
+      <version>1.33</version>
     </dependency>
     <dependency>
       <groupId>io.prometheus</groupId>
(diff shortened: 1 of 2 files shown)


matthiasr:
@or-shachar could you take a look at this to make sure I did it right?

or-shachar (approved):
`SafeConstructor` indeed sounds safer :) 

Important to note that this limits the kind of objects this constructor can return to standard java objects:
image ([from docs](https://bitbucket.org/snakeyaml/snakeyaml/wiki/Documentation#markdown-header-loading-yaml))


Also [this helped me understand the motivation a little further](https://sensei.securecodewarrior.com/recipes/scw:snakeyaml:safeconstructor-no-args).

The `CloudWatchCollectorTest` covers parsing yamls and loading configuration so I'm guessing this doesn't break anything.

Looks good to me :-)

---
#### future note:
Maybe we should explore defining a schema to the configuration and use type safe loading. I think this can be a good opportunity to use Java `record` sede. The gains would be a well defined configuration class with fields and more proper access to configuration fields. 

Also - moving to configuration schema can help with https://github.com/prometheus/cloudwatch_exporter/issues/484.

matthiasr:
I agree, a proper configuration schema would be much better: #512

PR #645, opened 2024-01-31
Support GCP Web Identity session credentials

connor-scio:


src/main/java/io/prometheus/cloudwatch/CloudWatchCollector.java
```diff
@@ -347,14 +357,53 @@ private void loadConfig(
...
+                  .setAudience(serviceAccountEmail)
+                  .setIncludeEmail(true)
+                  .build());
+      return idTokenResponse.getToken();
+    }
+  }
+
   private AwsCredentialsProvider getRoleCredentialProvider(Map<String, Object> config) {
+    String roleArn = (String) config.get("role_arn");
+    if (config.containsKey("assume_role_web_identity")) {
```
matthiasr, line 375: Since this only works for GCP, I think this config key should reflect it?

matthiasr (commented):
I _think_ I understand what this is about, but I am not sure. Could you please add a bit more description to the PR, as well as documentation of the configuration and how to use it to the README?

You don't have to implement other providers, but I would like to make sure we are not limiting future options: Can you briefly explain how we would extend this to support other providers' web identity integrations in the future?

matthiasr:
Also, please add the DCO sign-off to all your commits, it is a legal requirement for us to accept your contributions.

Issue #571, opened 2023-08-30
How to automatic fetch all the ALB (With Target Group) tags


JesusFrontelo:
When you have many Load Balancers, and each one has multiple target groups, its really hard to manually specify tags with group_left (tag_xxx, tag_yyy, tag_zzz) instead of doing it with group_right, which automatically joins all tags to the metric. 

Here i give you an example of my configuration to scrape metrics and a query, for use in graph or alerts. 

First of all, you have to define wich metrics you want to collect, and pay attention on the dimension of each metric. 

````
  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: ActiveConnectionCount
    aws_dimensions: [LoadBalancer]
    aws_tag_select:
      resource_type_selection: elasticloadbalancing:loadbalancer/app
      resource_id_dimension: LoadBalancer
    aws_statistics: [Sum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: ConsumedLCUs
    aws_dimensions: [LoadBalancer]
    aws_statistics: [Average, Maximum, Minimum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: ProcessedBytes
    aws_dimensions: [LoadBalancer]
    aws_statistics: [Sum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: NewConnectionCount
    aws_dimensions: [LoadBalancer]
    aws_statistics: [Sum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: RequestCount
    aws_dimensions: [LoadBalancer]
    aws_statistics: [Sum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: HealthyHostCount
    aws_dimensions: [TargetGroup, LoadBalancer]
    aws_statistics: [Average, Minimum, Maximum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: UnHealthyHostCount
    aws_dimensions: [TargetGroup, LoadBalancer]
    aws_tag_select:
      resource_type_selection: elasticloadbalancing:targetgroup
      resource_id_dimension: TargetGroup
    aws_statistics: [Average, Minimum, Maximum]

  - aws_namespace: AWS/ApplicationELB
    aws_metric_name: TargetResponseTime
    aws_dimensions: [TargetGroup, LoadBalancer]
    aws_statistics: [Average]
````
[...]

PR #648, opened 2024-02-08
fix: support aws_tag_select with non-standard resource id in metric

### LINKED ISSUE #268: ALB target group tag selections failing due to resource id
Hi. I fail to get any matching metrics with this configuration (running 0.8.0):
```yaml
    metrics:             
    - aws_namespace: AWS/ApplicationELB
      aws_metric_name: HealthyHostCount      
      aws_tag_select:                                
        tag_selections:             
          "ingress.k8s.aws/cluster": [myCluster]
          "kubernetes.io/namespace": [myNamespace]
        resource_type_selection: "elasticloadbalancing:targetgroup"
        resource_id_dimension: TargetGroup    
      aws_dimensions: [TargetGroup,LoadBalancer]       
      aws_statistics: [Minimum]
```

It works if I instead filter by LoadBalancer:
```yaml
        resource_type_selection: "elasticloadbalancing:loadbalancer"
        resource_id_dimension: LoadBalancer
```

From reading the source, it looks to me like the reason for this is that the resource IDs that are extracted from the ARNs don't match the dimension values returned by `listMetrics()` in the target group case.

The ResourceARNs returned by `aws resourcegroupstaggingapi get-resources ...` look like this for load balancers and target groups, respectively (lightly obfuscated):
```
"ResourceARN": "arn:aws:elasticloadbalancing:eu-north-1:1234567890:loadbalancer/app/ce9e96ed-somelbname-6b08/3a16156ee3244yyy"
"ResourceARN": "arn:aws:elasticloadbalancing:eu-north-1:1234567890:targetgroup/ce9e96ed-cbe56901daf6f7a1xxx/37def13b297a7yyy"
```

As far as I can tell, this would result in the following extracted resource IDs:
```
app/ce9e96ed-somelbname-6b08/3a16156ee3244yyy
ce9e96ed-cbe56901daf6f7a1xxx/37def13b297a7yyy
```

However, `aws list-metrics ...` returns these dimensions:
```
        {                                                                                                                                                                                                   
            "Namespace": "AWS/ApplicationELB",                                                                                                                                                              
            "MetricName": "HealthyHostCount",
            "Dimensions": [
                {
                    "Name": "TargetGroup",
                    "Value": "targetgroup/ce9e96ed-cbe56901daf6f7a1xxx/37def13b297a7yyy"
                },
                {
                    "Name": "LoadBalancer",
                    "Value": "app/ce9e96ed-somelbname-6b08/3a16156ee3244yyy"
                }
            ]
        }
```

So the value for the `LoadBalancer` dimension matches the extracted ID, but the one for `TargetGroup` doesn't (because of the `targetgroup/` prefix).

I can't say that I understand why the dimension values look like they do. Perhaps there's some CloudWatch bug there. But on the other hand it doesn't seem clear to me that the current resource ID extraction will work for all ARN flavours given the IMO less than super-clear documentation at https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html.

Could it be an option to just pass around the full ARNs in `CloudWatchCollector.java` and consider a metric a match as long as there's an ARN that _ends with_ the dimension value?

msvticket:
Fixes #268

@matthiasr  I'll add an update to the README when I get the word that the code is at least in the ball park ready to be merged.


matthiasr (commented):
Thank you for tackling this! I like giving users the option to supply patterns.

I see that we can both encode known patterns in `ARN_RESOURCE_ID_REGEXP_FOR_TYPE` _and_ allow configuring custom patterns. What guidance can we give future contributors on whether they should add patterns to the code or the examples?

Would it be "good enough" to keep the exporter oblivious of specific patterns, and only add the `elasticloadbalancing:targetgroup` to the examples?

matthiasr:
Will this replace the workaround outlined in #571 or is it separate?

msvticket:
> Thank you for tackling this! I like giving users the option to supply patterns.
> 
> I see that we can both encode known patterns in `ARN_RESOURCE_ID_REGEXP_FOR_TYPE` _and_ allow configuring custom patterns. What guidance can we give future contributors on whether they should add patterns to the code or the examples?
> 
> Would it be "good enough" to keep the exporter oblivious of specific patterns, and only add the `elasticloadbalancing:targetgroup` to the examples?

For me it would be good enough, but for the general user it would certainly be preferable for as a much as possible to work out of the box.

msvticket:
> Will this replace the workaround outlined in #571 or is it separate?

It would replace the label_replace trick.

matthiasr:
> general user it would certainly be preferable for as a much as possible to work out of the box

I'm afraid this exporter does not work "out of the box" at all 😬

one does not simply use cloudwatch exporter out of the box

The example configurations are the main way to "get started", and where we _do_ collect service specifics, so I think to keep things clear and maintainable going forward, I would rather remove `ARN_RESOURCE_ID_REGEXP_FOR_TYPE` and add it to the relevant example configurations instead.

matthiasr:
This has the added benefit of showing how to use it, so that hopefully users for other services can help themselves without having to wait for exporter code changes and releases.

matthiasr:
Awesome!

msvticket:
I have now removed ARN_RESOURCE_ID_REGEXP_FOR_TYPE from the code. I'll look at documentation soon

msvticket:
Now I have documented the new option and updated a couple of examples to use it

matthiasr:
Awesome, thank you! Merging despite the `build_image` failure – `build` succeeded, and the Docker version issue is resolved in master.

PR #661, opened 2024-03-03
Downgrade advanced-security/maven-dependency-submission-action action

matthiasr:
Due to advanced-security/maven-dependency-submission-action#65.


matthiasr:
@SuperQ FYI – we'll need to `@dependabot ignore` 4.0.1 when it tries to upgrade us again, until there's a fix for this

PR #684, opened 2024-04-15
Add DocumentDB Grafana dashboard

PLeS207:
Hello team, I was able to create Kubernetes Grafana dashboard for DocumentDB
Hope it will be a good example 
@matthiasr


matthiasr:
We don't have a DocumentDB or Grafana at hand, so I can't really test this, but it looks OK :) I'll merge it, and then we'll see if anyone wants to improve on it or add more dashboard examples.

PR #802, opened 2025-02-23
Update pom.xml to bump servlet vers

a-m-a-r-t-i-n:
Bumped servlet version to match https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/12.0.12


SuperQ:
Updated by dependabot.

PR #813, opened 2025-03-24
Update pom.xml with 2.31.6 awssdk version to address Netty CVE CVE-2025-24970

karthika-shastha:


SuperQ:
Updated by dependabot.

PR #817, opened 2025-04-22
making changes of openjdk

phoenix2501:


SuperQ:
We don't use CircleCI anymore.

PR #829, opened 2026-05-26
Replace CircleCI with Github Actions

ArthurSens:


.github/workflows/maven.yml
```diff
@@ -1,35 +1,122 @@
...
+          username: ${{ secrets.QUAY_IO_LOGIN }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+      - name: Login to Docker Hub
+        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4.0.0
+        with:
+          username: ${{ secrets.DOCKER_HUB_LOGIN }}
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+      - name: Build and push image
+        run: |
+          tags=(master main)
```
ArthurSens: This was on the CircleCI config, but I don't think we need to publish two tags, do we? 🤔

SuperQ: This was added in https://github.com/prometheus/cloudwatch_exporter/pull/378. I don't think it's necessary. Let's just use the standard publish pipeline here.

SuperQ:
Disabled in circleci.

PR #830, opened 2026-05-28
build: add distroless image variant

roidelapluie:
The distroless image runs as nonroot (UID 65532) without a shell, reducing the attack surface compared to the eclipse-temurin-based image.

### DIFF
--- a/Dockerfile.distroless
+++ b/Dockerfile.distroless
@@ -1,0 +1,36 @@
+FROM eclipse-temurin:21-jdk-noble AS builder
+
+SHELL ["/bin/bash", "-xe", "-o", "pipefail", "-c"]
+
+ARG MAVEN_VERSION=3.8.5
+ARG MAVEN_SHA512=89ab8ece99292476447ef6a6800d9842bbb60787b9b8a45c103aa61d2f205a971d8c3ddfb8b03e514455b4173602bd015e82958c0b3ddc1728a57126f773c743
+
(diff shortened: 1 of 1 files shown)


SuperQ (approved):

SuperQ:
I don't think this will actually build/publish since this repo doesn't use the common Makefile due to Java.

Dockerfile.distroless
```diff
@@ -0,0 +1,36 @@
+FROM eclipse-temurin:21-jdk-noble AS builder
```
SuperQ, line 1: FYI, I've updated to Java 25.

Dockerfile.distroless
```diff
@@ -0,0 +1,36 @@
...
+ && tar -x --strip-components=1 -C /opt/maven -f /opt/maven.tar.gz
+ENV PATH=/opt/maven/bin:${PATH}
+
+WORKDIR /cloudwatch_exporter
+COPY . /cloudwatch_exporter
+
+RUN mvn package \
+ && mv target/cloudwatch_exporter-*-with-dependencies.jar /cloudwatch_exporter.jar
+
+FROM gcr.io/distroless/java21-debian12:nonroot
```
loosebazooka, line 20: fyi debian12 images are unmaintained, please use a debian13 image

loosebazooka: see: https://github.com/GoogleContainerTools/distroless/#what-images-are-available

PR #846, opened 2026-06-05
Update build deps

SuperQ:
* Pin GitHub actions to SHAs.
* Bump Maven version.
* Update dependabot to ignore synced files.


ArthurSens (approved):
I wonder if we could add zizmor to the repo sync automation as well. Zizmor would enforce the pinning everywhere

PR #852, opened 2026-06-05
chore: Added Maven wrapper for consistent build environment

dhoard:
Added Maven wrapper (mvnw)


SuperQ (commented):
Can you explain a little "why" for this change?

dhoard:
It's a common practice to support a consistent toolchain/build environment and not require Maven to be installed.

SuperQ (approved):

PR #855, opened 2026-06-05
chore: Added/updated unit tests to validate code

dhoard:
- Updated code to use JUnit 6
- Updated code to use AssertJ
- Added unit tests for code coverage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea/
 /targetpom.xml.releaseBackup
 release.properties
 target/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 /targetpom.xml.releaseBackup
 release.properties
 target/
--- a/scripts/stress-test.sh
+++ b/scripts/stress-test.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright (c) 2026-present Douglas Hoard
+# Copyright (c) 2026 The Prometheus Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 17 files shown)


scripts/stress-test.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2026-present Douglas Hoard
```
SuperQ: Hmm, typically we assign copyright to `The Prometheus Authors`.

If this is sourced / vendored from another repo we need to have it listed in a `NOTICE` file.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
SuperQ: Please don't include your local editor ignores in the repo. Use [`git config --global core.excludesFile`](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).


```suggestion
```

SuperQ (changes requested):

scripts/stress-test.sh
dhoard: My Mistake. Will resolve.

dhoard: Fixed.

.gitignore
dhoard: Fixed.

scripts/stress-test.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2026 The Prometheus Authors
```
SuperQ:
```suggestion
# Copyright (c) The Prometheus Authors
```

SuperQ (approved):

PR #873, opened 2026-07-01
Fixup GitHub configs

SuperQ:
* The dependabot config filename should be `dependabot.yml`.
* Rename workflow from build to ci to match ecosystem naming convention.


dhoard (approved):

SuperQ:
Added a change to avoid overwrites by the sync tool: https://github.com/prometheus/prometheus/pull/19051

PR #882, opened 2026-07-12
Add workflow that allow triagers to approve workflows

ArthurSens:
Based on https://github.com/prometheus/promci/pull/114


SuperQ:
Maybe we should add this to the sync script in prometheus/prometheus?

SuperQ:
Completed in https://github.com/prometheus/cloudwatch_exporter/pull/883

